PR #1, opened 2016-04-08
removed a useless pmd rule

ggraffieti:


DanySK:
Better use simple present in commit messages.

PR #7, opened 2016-09-04
Alchemist effects in human-readable format

w4bo:
Effects are stored in human readable format.

I added a serialization factory to centralize effects serialization. Future changes in serialization will affect only this class.

I had to include RuntimeTypeAdapterFactory.java from the [Gson repo](https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java) as this class is not shipped with the Gson artifact.


DanySK:
Including that file is a no-go. We have not the power to maintain our own pieces of Gson. If you need a newer version of Gson, then please check if it's already there in Maven Central

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/JEffectsTab.java
```diff
@@ -263,13 +264,13 @@ public void setEffects(final List<Effect> effects) {
...
-        moveLeftButton.setEnabled(value);
-        moveRightButton.setEnabled(value);
-        for (final Component c : stackSec.getComponents()) {
+        this.addEffectButton.setEnabled(value);
+        this.remEffectButton.setEnabled(value);
+        this.saveButton.setEnabled(value);
+        this.loadButton.setEnabled(value);
+        this.moveLeftButton.setEnabled(value);
+        this.moveRightButton.setEnabled(value);
+        for (final Component c : this.stackSec.getComponents()) {
```
DanySK: Not  a problem - but why this "this" invasion? Where they necessary, or was it the setting of your editor? If so, be wary: in many cases, abusing this. leads to unnecessary hyperverbose code (and java is already verbose itselft)

w4bo: My bad. I am used to using this setting. I will get rid of "this" invasion

PR #11, opened 2017-02-21
fixed broken JSON loading and clarified gps trace obtaining from node

gscaparrotti:


alchemist/alchemist-maps/src/main/java/it/unibo/alchemist/model/implementations/movestrategies/speed/TraceDependantSpeed.java
```diff
@@ -37,7 +38,7 @@ public TraceDependantSpeed(final MapEnvironment<T> e, final Node<T> n, final Rea
         env = e;
         node = n;
         reaction = r;
```
DanySK: Could you requireNonNull also on other input parameters?

alchemist/alchemist-maps/src/main/java/at/jku/traces/json/JKUJSONLoader.java
```diff
@@ -23,15 +24,33 @@
...
  */
 public final class JKUJSONLoader implements Serializable {
 
     private static final long serialVersionUID = 7144531714361675479L;
-    private static final Gson GSON = new Gson();
+    private static final Gson GSON = new GsonBuilder()
+            .registerTypeAdapter(GPSPoint.class, new JsonDeserializer<GPSPoint>() {
+                @Override
+                public GPSPoint deserialize(final JsonElement json, final Type typeOfT, final JsonDeserializationContext context) throws JsonParseException {
+                    final JsonObject obj = json.getAsJsonObject();
```
DanySK: Let's check that it is an object first (same with the doubles), and fail by IllegalStateException otherwise, so that in case of future problems we've got clear indications.

alchemist/alchemist-maps/src/main/java/it/unibo/alchemist/model/implementations/movestrategies/speed/TraceDependantSpeed.java
gscaparrotti: The bare objects are already checked for null values by calling LangUtils.requireNonNull(e, n, r). Do you prefer to change it and call requireNonNull the same way it is now done in the new line? I don't think it would make that much difference.

DanySK: Nah, it's fine as it is. I was fooled by the diff, my fault.

alchemist/alchemist-maps/src/main/java/at/jku/traces/json/JKUJSONLoader.java
```diff
@@ -23,15 +24,41 @@
...
 
     private static final long serialVersionUID = 7144531714361675479L;
-    private static final Gson GSON = new Gson();
+    private static final Gson GSON = new GsonBuilder()
+            .registerTypeAdapter(GPSPoint.class, new JsonDeserializer<GPSPoint>() {
+                @Override
+                public GPSPoint deserialize(final JsonElement json, final Type typeOfT, final JsonDeserializationContext context) throws JsonParseException {
+                    if (json.isJsonObject()) {
+                        final JsonObject obj = json.getAsJsonObject();
+                        try {
```
DanySK: Don't use exceptions as control flow: check that these objects are doubles before calling the getter

alchemist/alchemist-maps/src/main/java/at/jku/traces/json/JKUJSONLoader.java
```diff
@@ -23,15 +24,41 @@
...
+                            final double latitude = obj.get("la").getAsDouble();
+                            final double longitude = obj.get("lo").getAsDouble();
+                            final double time = obj.get("t").getAsDouble();
+                            return new GPSPointImpl(latitude, longitude, time); 
+                        } catch (final IllegalStateException | ClassCastException e) {
+                            throw new IllegalStateException("An invalid JSON has been provided: unable to get the double values");
+                        }
+                    } else {
```
DanySK: Remove redundant else

alchemist/alchemist-maps/src/main/java/at/jku/traces/json/JKUJSONLoader.java
DanySK: (btw, [if you want to deepen why so](https://shipilev.net/blog/2014/exceptional-performance/))

PR #25, opened 2017-04-24
fixed antrl versions conflict

gscaparrotti:
Conflicts:
	alchemist/gradle.properties


DanySK:
Let's see if the build goes through

PR #35, opened 2017-05-31
Solve two serious bugs in RunScafiProgram action + bottleneck

metaphori:
Bugs solved
1. One related to sending exports to neighbours (due to shadowing of vars)
2. The other related to the lack of information, in particular sensors, about the node itself, due to bad initialization check


metaphori:
Solved one serious bottleneck at the startup, namely the runtime evaluation of Scala expressions for concentration values, via caching

DanySK:
I am going to switch this to LoadingCache, it's better than having to deal with caching manually

PR #43, opened 2017-06-29
Scafi incarnation fix

metaphori:
* Updates root build.gradle to run ScalaTest tests
* Fixes a bug in scafi incarnation related to molecule concentration setting on YAML loading


metaphori:
gradle-scalatest plugins seems to **replace** the "test" task so it uncovers JUnit tests. Go for another approach

Issue #51, opened 2017-10-19
Files are brutally deleted when create a new project 


Pptr95:
When you try to create e new project and select a not empty folder, all the files within that folder are brutally deleted.

DanySK:
@Pptr95, can I mark this solved?

PR #52, opened 2017-10-24
Develop - Bug #51 fixed

Pptr95:


alchemist/alchemist-projectview/src/main/resources/it/unibo/alchemist/l10n/ProjectViewUIStrings.properties
```diff
@@ -127,7 +127,11 @@ select_folder = Select folder
 select_folder_proj = Select folder project
 select_folder_full = Folder not empty
 select_folder_full_header = This selected folder is not empty
-select_folder_full_content = To create a new project, you must choose an empty folder. Do you want to initialize this folder as an empty folder?
+<<<<<<< HEAD
```
DanySK: Nope, that is an unresolved merge conflict.

DanySK:
And please use better commit messages. There are three consecutive, meaningless "Bug fixed" messages! Which bug? How was it fixed?

PR #53, opened 2017-10-25
Fixed bug #51

Pptr95:
Now is possible to create new projects within not empty folders without remove all file inside


DanySK:
Good to go

DanySK:
Good to go... but please reopen it against `develop`

PR #62, opened 2018-03-27
Update scafi incarnation for scafi v0.3.0

metaphori:


alchemist/alchemist-incarnation-scafi/build.gradle
```diff
@@ -4,7 +4,7 @@ dependencies {
     compile (
         "com.github.cb372:scalacache-core_${scalaMajorVersion}:${scalaCacheVersion}",
         "com.github.cb372:scalacache-guava_${scalaMajorVersion}:${scalaCacheVersion}",
-        "it.unibo.apice.scafiteam:scafi-core_${scalaMajorVersion}:${scafiVersion}",
+        "it.unibo.apice.scafiteam:scafi-lib_${scalaMajorVersion}:${scafiVersion}",
```
DanySK: I guess this change is because `-lib` pulls in `-core`, right? I'm a bit torn about whether or not also include core, mostly for clarity. Not a problem though.

DanySK:
Nice one, good to go!

alchemist/alchemist-incarnation-scafi/build.gradle
metaphori: Indeed.

PR #78, opened 2018-11-14
Update with develop branch, fix simulation crashing when pausing

### LINKED ISSUE #59: Add support to Select and Pan modes
The new JavaFX UI supports switching between Select Mode and Pan Mode, but nothing changes.

The FXOutputMonitor classes should handle the DragEvents to make these modes work.

Vuksaa:
Also, should close #59.


DanySK:
Let's do this step

PR #86, opened 2018-12-10
Write companion object for keybind management and write new GUI for rebinding keys

Vuksaa:


alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/GenericEvents.kt
```diff
@@ -58,19 +58,19 @@ abstract class AbstractEventDispatcher<T : TriggerAction, E : Event> : EventDisp
 
     protected var triggers: Map<T, (event: E) -> Unit> = emptyMap()
```
DanySK: Do we need protected?

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/KeyboardEvents.kt
```diff
@@ -17,7 +17,9 @@ import javafx.scene.input.KeyEvent
  */
 enum class ActionOnKey {
     PRESSED,
-    RELEASED
+    RELEASED;
+
+    operator fun plus(key: KeyCode): KeyboardTriggerAction = KeyboardTriggerAction(this, key)
```
DanySK: uhmm... If understand correctly, this combines an `ActionOnKey` and a `KeyCode` to a `TriggerAction`. I'm not sure we want to use `+`, it's not a natural sum. maybe an infix function `with`?

``keyboard.setOnAction(ActionOnKey.PRESSED with it, enqueueMove)``

Or another keyword, I'm open to suggestions, I'ma bit reluctant to overload operators too much.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/Keybinds.kt
```diff
@@ -32,7 +32,7 @@ enum class ActionFromKey {
 class Keybinds {
     companion object {
         private val filesystemPath = "${System.getProperty("user.home")}${File.separator}Alchemist${File.separator}"
-        private const val classpathPath = "it/unibo/alchemist/gui/"
+        private val classpathPath = "it${File.pathSeparator}unibo${File.pathSeparator}alchemist${File.pathSeparator}gui${File.pathSeparator}"
```
DanySK: Does it access a file, or does it load from classpath? In the latter case, `/` is to be used. The classpath is always separated by `/`

alchemist/alchemist-swingui/src/main/resources/it/unibo/alchemist/gui/keybinds.json
```diff
@@ -0,0 +1,33 @@
...
+  "ONE_STEP": [
+    "PERIOD"
+  ],
+  "PAN_EAST": [
+    "D"
+  ],
+  "MODIFIER_SHIFT": [
+    "SHIFT"
+  ]
+}
```
DanySK: super nice

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/Keybinds.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        /**
+         * Retrieve the keys bound to a certain action.
+         */
+        operator fun get(action: ActionFromKey): Set<KeyCode> = config[action] ?: emptySet()
+
+        /**
+         * Associate an action with a set of keys.
+         */
+        operator fun set(action: ActionFromKey, keys: Set<KeyCode>) { config += action to HashSet(keys) }
```
DanySK: Can this get overloaded by:
``operator fun set(action: ActionFromKey, key: KeyCode) { config += action to setOf(keys) }``?

so that we get rid of all the `setOf` calls in the default binding

Moreover, why to we call `HashSet(keys)`? Can't we just pass down the `Set`? It's read only (and covariant) anyway.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/GenericEvents.kt
Vuksaa: Protected is not necessary, but I preferred to create the "set" operator to add triggers instead of making the map public since I did not intend the triggers to be removable once added.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/KeyboardEvents.kt
Vuksaa: That makes sense, I made the adjustment.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/Keybinds.kt
Vuksaa: That version of Keybinds.kt is outdated, in a later commit I fixed this issue.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/Keybinds.kt
Vuksaa: That version of Keybinds.kt is outdated, in a later commit I fixed this issue.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/GenericEvents.kt
DanySK: Not very clear to me. If it's initialized once and then hidden as an implementation detail, why it's not a `private lateinit`?

DanySK:
@Vuksaa, please pull from my develop and push on your working branch, I updated the build to publish a scan also on pull requests, to help debug them.

alchemist/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/input/GenericEvents.kt
Vuksaa: Now I remember why I made it specifically `protected`. I wanted to allow the inheriting classes to manage the triggers. It's not `lateinit ` for that very reason; this way the triggers can be accessed at any time by the inheriting classes without having to worry whether the map has been initialized or not.

DanySK: OK, then add a documentation comment clearly explaining how subclasses are supposed to inherit it, which is the expected usage.

Thanks

Vuksaa: It is done

DanySK:
OK. Next issue: it does not work on Java 11 and Java 12.
``
Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    /home/travis/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-reflect/1.3.0/6fd129fd9ba8581f2cb9c58bfd431dda4ee0457e/kotlin-reflect-1.3.0.jar (version 1.3)
    /home/travis/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-jdk8/1.2.60/7d5d50242f2bd1bcb6663044fbc814e0f59255aa/kotlin-stdlib-jdk8-1.2.60.jar (version 1.2)
    /home/travis/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-jdk7/1.2.60/31f1ea280ab4942ed25f91760d7170c41f78d06f/kotlin-stdlib-jdk7-1.2.60.jar (version 1.2)
    /home/travis/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/1.3.0/a134b0cfe9bb44f98b0b3e889cda07923eea9428/kotlin-stdlib-1.3.0.jar (version 1.3)
    /home/travis/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-common/1.3.0/84a2e0288dc17cd64d692eb1e5e0de8cd5ff0846/kotlin-stdlib-common-1.3.0.jar (version 1.3)
w: Some runtime JAR files in the classpath have an incompatible version. Consider removing them from the classpath
``

@Vuksaa are you able to check dependencies in Gradle and understand why this happens? How comfortable are you with Gradle?

Vuksaa:
I believe the issue here is that we are using version 1.3 of kotlin-stdlib but tornadofx 1.7.17 used version 1.2. I updated tornadofx to version 1.7.18 which uses kotlin-stdlib 1.3.
The travis build should still fail on java 11 and 12 though, because tornadofx does not run on java 9 or above yet.

DanySK:
Then unfortunately this cannot get merged until TornadoFX will get to be Java 11 compatible. Moreover, we will probably need separate build procedures (part of them are in place already) to support separate releases for JDK8 and JDK11+.
We need edvin/tornadofx#899 to get addressed. We might - but I'd like to avoid it if possible - fork edvin/tornadofx and release a JDK11 compatible beta version under a different group, and use it to check if our project works as expected.
I asked @edvin for a rough estimate of the release time, let's compare it to ours and choose the strategy to adopt.

edvin:
Maybe you could contribute any changes back to the Java 11 version of TornadoFX and thus help get the ball rolling on the official release?

DanySK:
Of course. Any change in my fork will be contributed upstream.
Cheers

edvin:
Perfect! I'm here if you need anything at all.

DanySK:
We got a problem:
`warning: /home/travis/.gradle/caches/modules-2/files-2.1/org.controlsfx/controlsfx/9.0.0/8e009b653a2a6eb3336016ad314c040e72395914/controlsfx-9.0.0.jar(org/controlsfx/control/PopOver$ArrowLocation.class): major version 53 is newer than 52, the highest major version supported by this compiler.`

We must use an older controlsfx in the Java 8 build, and a newer one in the Java 11 (if necessary, but I guess it will be).

Vuksaa:
There's also another issue.

`MapWormhole#setEnvPosition` has been changed to use `GeoPosition#getLatitude` and `GeoPosition#getLongitude`, but in my fork setEnvPosition accepts a `Position2D<?>` rather than a `GeoPosition` so it obviously doesn't work as of right now.
One possible fix would be to check the parameter with `instanceof` and cast it to a `GeoPosition`, but I understand that the preferred method would be to have the parameter be a `GeoPosition`.

And that brings up a problem I noticed when doing this last merge conflict resolution: the root interfaces [IWormhole2D](https://github.com/AlchemistSimulator/Alchemist/blob/develop/alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/wormhole/interfaces/IWormhole2D.java) (main repo) and [BidimensionalWormhole](https://github.com/Vuksaa/Alchemist/blob/develop/alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/wormhole/interfaces/BidimensionalWormhole.java) (my repo) are different.
In the main repo, `getViewPoint` and `setEnvPosition` accept a `P` but in my repo they accept a `Position2D<?>`; this is the why we can't change `MapWormhole#setEnvPosition` to accept a `GeoPosition`.

The reason we can't just make `getViewPoint` and `setEnvPosition` accept a `P` in my version is that it would break functions such as [`DrawDot#consumeData`](https://github.com/Vuksaa/Alchemist/blob/8d88de6991acc260b53dffc9108c3e0af650bb57/alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawDot.java#L85) because the erasure doesn't allow us to pass a `Position2D<?>` to a `BidimensionalWormhole<?>#getViewPoint`. They would have to be of the same generic type, right?

So I came up with a dirty fix: make `consumeData` parameterized and adjust the code like so:
```
@Override
protected <P extends Position2D<? extends P>> Queue<DrawCommand> consumeData() {
    final double size = getSize();
    return positions.stream().<DrawCommand>map((Position2D<?> position) -> (GraphicsContext graphic, BidimensionalWormhole<?> wormhole) -> {
        final BidimensionalWormhole<P> wh = (BidimensionalWormhole<P>) wormhole;
        final Point viewPoint = wh.getViewPoint((P) position);
        final double startX = viewPoint.getX() - size / 2;
        final double startY = viewPoint.getY() - size / 2;
        graphic.setFill(getColor());
        graphic.fillOval((int) startX, (int) startY, (int) size, (int) size);
    }).collect(Collectors.toCollection(Queues::newConcurrentLinkedQueue));
}
```
But this time we get a warning for unchecked cast. We cannot use `instanceof` here because erasure happens. So we would have to handle the exceptions by wrapping the `DrawCommand`s' functions in try-catch blocks.

Is there a clean way to fix this? I feel like my solution is a cheap trick more than anything. How should we handle this?

DanySK:
As per our conversation, let me know how this progresses.

.gitignore
```diff
@@ -1,19 +1,17 @@
...
+version.info
+prepare_environment.sh
+**/*-gen/
+**/*generated*/
+*.cache-main
+*.cache-tests
+out/
+.DS_Store
+*.kate-swp
+**/*.kotlintest*/*
```
DanySK: @Vuksaa why is this file rewritten?

.travis.yml
```diff
@@ -1,24 +1,80 @@
...
+os:
+ - linux
+ - osx
+dist: xenial
 language: java
-before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -y graphviz
+jdk:
+  - openjdk8
```
DanySK: @Vuksaa, kill OpenJDK8. Let's make sure this works correctly on JDK11 and above.

DanySK:
@Vuksaa it's time we merge to the mainline. We need to:
* Drop Java 8 support (do not test for it on the CI)
* Make sure it compiles and runs tests on the CI

DanySK:
Recover the deploy section, in there just change openjdk8 to openjdk11. Build matrix is fine.

DanySK:
There is still a compilation issue, though

PR #90, opened 2019-03-18
Fix Overlap Tests in Biochemistry Incarnations

giuluck:
Fixed test testNoOverlapInSimulation1 in TestBioRect2DEnvironmentNoOverlap.java, with minor refactoring of other tests due to internal changes.

### DIFF
--- a/alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
+++ b/alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
@@ -40,7 +40,13 @@ import static org.junit.Assert.*;
  */
 public class TestBioRect2DEnvironmentNoOverlap {
 
-    private static final double DELTA = 1e-6;
+    /**
+     * The test seems to work for values of delta up to 2*10^(-8)
+     *
+     * As future changes in the core can vary the sensibility of floating point operations, leading to an undesired
+     * test failure, the value has been set to the next order of magnitude.
+     */
+    private static final double DELTA = 1e-7;
     private static final double BIG_CELL_DIAMETER = 30;
     private static final double MEDIUM_CELL_DIAMETER = 20;
     private static final double LITTLE_CELL_DIAMETER = 10;
--- a/alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
+++ b/alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
@@ -8,15 +8,11 @@
 package it.unibo.alchemist.test;
 
 import java.io.InputStream;
-import java.util.AbstractMap;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.apache.commons.math3.util.FastMath;
-import org.apache.commons.math3.util.Pair;
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.kaikikm.threadresloader.ResourceLoader;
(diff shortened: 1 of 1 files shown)


alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
```diff
@@ -714,39 +698,37 @@ public void finished(final Environment<Double, Euclidean2DPosition> env, final T
...
-                            System.out.println("nodes: ");
-                            System.out.println("center : " + env.getPosition(n));
-                            listOverlapping.forEach(c -> System.out.println("In range : " + env.getPosition(c)));
-                            listOverlapping.forEach(c -> System.out.println("distance : " + env.getPosition(c).getDistanceTo(env.getPosition(n))));
-                        }
-                    });
-                    return posResult;
-                }
+                        .map(n -> (CellWithCircularArea<Euclidean2DPosition>) n)
+                        .flatMap(n -> env.getNodesWithinRange(n, n.getDiameter() - DELTA).stream().peek(c -> {
```
DanySK: Peeking on a stream is very, very non functional. Try to collect the information, and fail only on termination:
```java
n -> env.getNodesWithinRange(n, n.getDiameter() - DELTA).stream()
    .map(n -> new Pair(c, n))
```

alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
```diff
@@ -714,39 +698,37 @@ public void finished(final Environment<Double, Euclidean2DPosition> env, final T
...
+                            fail("Nodes " + n.getId() + env.getPosition(n) + " and " + c.getId() + env.getPosition(c) +
+                                    " are overlapping. Their distance is: " + env.getDistanceBetweenNodes(n, c));
+                        }))
+                        .findAny()
```
DanySK: With the proposed change, you can here do something like:
```java
.findAny()
.ifPresent(it -> fail(...))
```

alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
```diff
@@ -39,12 +31,14 @@
...
 
+import static org.junit.Assert.*;
+
 /**
  *
  */
 public class TestBioRect2DEnvironmentNoOverlap {
 
-    private static final double MAX_PRECISION = 0.9999999999;
+    private static final double DELTA = 1e-6;
```
DanySK: Looks rather arbitrary: up to which value does the test work? What happens if you put a zero? What if you put `nextUp(0)`?

giuluck: The test works for values up to DELTA=1e-7 and starts failing with DELTA=1e-8 (same for DELTA=0, obviously, but giving in the error log that two nodes with distance d=1.0 are overlapping, which should be acceptable so a DELTA is needed).
Eventually, as nextUp(0) returns a value which is about 1e-45, the test wil fail too.

alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
```diff
@@ -782,7 +764,7 @@ private boolean thereIsOverlap(final Environment<Double, Euclidean2DPosition> en
                                 .filter(c -> env.getDistanceBetweenNodes(c, n)
                                         < ((((CellWithCircularArea<Euclidean2DPosition>) n).getRadius()
```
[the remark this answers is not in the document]
giuluck: This was an old piece of code I didn't change.
In the last commit, I made a general refactoring of those two tests removing code duplications and fixing the code style.

alchemist/alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBioRect2DEnvironmentNoOverlap.java
DanySK: I'd say here documenting (with a comment in code) which is the limit at which the test works is fine.

giuluck: Now it should be ok

PR #117, opened 2019-11-07
More adjustments to scafi incarnation

metaphori:


alchemist/alchemist-incarnation-scafi/src/main/scala/it/unibo/alchemist/model/implementations/nodes/NodeManager.scala
```diff
@@ -13,13 +13,16 @@ import it.unibo.alchemist.model.interfaces.{Node}
...
 
 class SimpleNodeManager[T](val node: Node[T]) extends NodeManager {
   override def put[V](molecule: String, concentration: V): Unit = node.setConcentration(new SimpleMolecule(molecule), concentration.asInstanceOf[T])
 
   override def get[V](molecule: String): V = node.getConcentration(new SimpleMolecule(molecule)).asInstanceOf[V]
 
-  override def has(molecule: String): Unit = node.contains(new SimpleMolecule(molecule))
+  override def has(molecule: String): Boolean = node.contains(new SimpleMolecule(molecule))
+
+  override def remove(molecule: String): Unit = node.removeConcentration(new SimpleMolecule(molecule))
```
metaphori: @DanySK would it be better to cache the molecules?

DanySK: A `Map<String, Molecule>` would be inefficient, molecules are optimized to be faster to lookup than strings.
Your proposal sounds to me as premature optimization, and as you know that's the root of all evil.
However, if I recall the implementation correctly, it may make sense to adopt a Flyweight pattern for the simplest molecules (basically all but SAPERE's) which, besides saving memory, might allow for faster comparison. Let's move such discussion to #118

PR #121, opened 2019-11-27
Add effects for drawing layers

paagamelo:
### Two effects were added

- one for drawing layers values with _isolines_:

isolines

- one for drawing layers values with _gradients_:
gradient

### The code
New code is quite a lot, this because a whole new package was added: the isolines package. Such package contains basic astractions and algorithms to find isolines, and is meant to be used with the new gui too (ideally, without modifying it at all). Apart from that, a few classes for  the new effects were added too. Every choice is explained in the documentation. Plus, may it be helpful, I will attach two UML diagrams: one for the new isolines package and one describing the new effect classes hierarchy.

Isolines
DrawLayers


alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
```diff
@@ -0,0 +1,292 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @SuppressWarnings({"PMD.CompareObjectsWithEquals", "unchecked"})
+    @SuppressFBWarnings("ES_COMPARING_STRINGS_WITH_EQ")
+    @Override
+    public <T, P extends Position2D<P>> void apply(final Graphics2D g, final Node<T> n, final Environment<T, P> env, final IWormhole2D<P> wormhole) {
+        // if marker node is no longer in the environment or it is no longer displayed, we need to change it
+        if (markerNode.isPresent() && (!env.getNodes().contains(markerNode.get())
+                                        || !wormhole.isInsideView(wormhole.getViewPoint(env.getPosition((Node<T>) markerNode.get()))))) {
```
DanySK: Why so indented?

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
```diff
@@ -0,0 +1,292 @@
...
+                 */
+                final Thread th = new Thread(() -> molecule = incarnation.createMolecule(molString));
+                th.start();
+                try {
+                    th.join();
+                } catch (final InterruptedException e) {
+                    L.error("Bug.", e);
+                }
+            }
+
```
DanySK: No empty lines. If there is a separator, then write a comment. Otherwise, compact the code.

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
```diff
@@ -0,0 +1,292 @@
...
+        }
+        if (markerNode.isEmpty()) {
+            markerNode = Optional.of(n);
+        }
+        if (markerNode.get() == n) { // at this point markerNode.isPresent() is always true, so we directly get it
+            if (layerFilter && (incarnation == null || curIncarnation != prevIncarnation || molString != molStringCached)) {
+                molStringCached = molString;
+                prevIncarnation = curIncarnation;
+                incarnation = SupportedIncarnations.get(curIncarnation.getCurrent())
+                        .orElseThrow(() -> new IllegalStateException(curIncarnation.getCurrent() + " is not a valid incarnation."));
```
DanySK: This is an horrible kludge and looks like it's copypasted from elsewhere.
Avoid copy and paste, we already have a zillion lines of code to maintain, so every time you feel like copying something, either use inheritance or composition (strategy).

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
```diff
@@ -0,0 +1,292 @@
...
+            colorCache = new Color(red.getVal(), green.getVal(), blue.getVal(), alpha.getVal());
+            g.setColor(colorCache);
+
+            final List<Layer<T, P>> toDraw = new ArrayList<>();
+            if (layerFilter && molecule != null && env.getLayer(molecule).isPresent()) {
+                toDraw.add(env.getLayer(molecule).get());
+            } else {
+                toDraw.addAll(env.getLayers());
+            }
+
```
DanySK: space

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
```diff
@@ -0,0 +1,292 @@
...
+                try {
+                    th.join();
+                } catch (final InterruptedException e) {
+                    L.error("Bug.", e);
+                }
+            }
+
+            colorCache = new Color(red.getVal(), green.getVal(), blue.getVal(), alpha.getVal());
+            g.setColor(colorCache);
+
```
DanySK: space

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayersGradient.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected <T, P extends Position2D<P>> void drawValues(final Function<? super P, ? extends Number> f, final Environment<T, P> env, final Graphics2D g, final IWormhole2D<P> wormhole) {
+        if (minOrMaxLayerValuesNeedsToBeUpdated()) {
+            updateMinAndMaxLayerValues();
+        }
+
```
DanySK: fix spacing

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayersIsolines.java
```diff
@@ -0,0 +1,234 @@
...
+ */
+public abstract class DrawLayersIsolines extends DrawLayersValues {
+
+    private static final int MAX_NUMBER_OF_ISOLINES = 50;
+    private static final long serialVersionUID = 1L;
+    @ExportForGUI(nameToExport = "Number of isolines")
+    private RangedInteger nOfIsolines = new RangedInteger(1, MAX_NUMBER_OF_ISOLINES, MAX_NUMBER_OF_ISOLINES / 4);
+    @ExportForGUI(nameToExport = "Distribution between min and max")
+    private Distribution distribution = Distribution.LINEAR;
+
```
DanySK: space

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayersIsolines.java
```diff
@@ -0,0 +1,234 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    protected abstract <T, P extends Position2D<P>> void drawLayers(Collection<Layer<T, P>> toDraw, Environment<T, P> env, Graphics2D g, IWormhole2D<P> wormhole);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected <T, P extends Position2D<P>> void drawValues(final Function<? super P, ? extends Number> f, final Environment<T, P> env, final Graphics2D g, final IWormhole2D<P> wormhole) {
```
DanySK: spacing

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayersIsolines.java
```diff
@@ -0,0 +1,234 @@
...
+
+        if (nOfIsolinesCached != nOfIsolines.getVal()
+                || minOrMaxLayerValuesNeedsToBeUpdated()
+                || distributionCached != distribution) {
+            nOfIsolinesCached = nOfIsolines.getVal();
+            updateMinAndMaxLayerValues();
+            distributionCached = distribution;
+
+            final double[] l;
+            if (distribution == Distribution.LOGARITHMIC) {
```
DanySK: ternary operator

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayersValues.java
```diff
@@ -0,0 +1,176 @@
...
+        return !minLayerValueCached.equals(minLayerValue) || !maxLayerValueCached.equals(maxLayerValue);
+    }
+
+    /**
+     * Update min and max layer values.
+     */
+    protected void updateMinAndMaxLayerValues() {
+        minLayerValueCached = minLayerValue;
+        maxLayerValueCached = maxLayerValue;
+
```
DanySK: space

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/ConcreteIsolinesFactory.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Segment2D makeSegment(final Number x1, final Number y1, final Number x2, final Number y2) {
+        Objects.requireNonNull(x1);
+        Objects.requireNonNull(y1);
+        Objects.requireNonNull(x2);
+        Objects.requireNonNull(y2);
+
```
DanySK: spacing

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/Utilities.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Checks if a variable number of objects are null, in that case an IllegalArgumentException will be thrown.
+     * @throws IllegalArgumentException
+     *              with "null argument" message
+     * @param objects
+     *              the objects to check
+     */
+    public static void requireNonNull(final Object...objects) {
```
DanySK: What an overkill... where is this used?

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/conrec/Conrec.java
```diff
@@ -0,0 +1,269 @@
...
+    private static final int CASE_1 = 1;
+    private static final int CASE_2 = 2;
+    private static final int CASE_3 = 3;
+    private static final int CASE_4 = 4;
+    private static final int CASE_5 = 5;
+    private static final int CASE_6 = 6;
+    private static final int CASE_7 = 7;
+    private static final int CASE_8 = 8;
+    private static final int CASE_9 = 9;
+    private static final int[][][] CASTAB =
```
DanySK: WTF is this?

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/conrec/Conrec.java
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Conrec a straightforward method of contouring some surface represented a regular
+ * triangular mesh.
+ *
+ * Ported from the C++ code by Nicholas Yue (see above copyright notice).
+ * See <a href="http://paulbourke.net/papers/conrec">http://paulbourke.net/papers/conrec</a> for full description
+ * of code and original C++ source.
+ *
+ * author  Bradley White
```
DanySK: Ugh

Super nope.

What's the license of this code? We can't use random code from the Internet, is this GPL-compatible?

Plus, copy-pasting code inside a project is **very** bad. Is there any library with this functionality?

DanySK:
Also, pay attention to duplication: https://app.codacy.com/manual/Alchemist/Alchemist/pullRequest?bid=15379231&prid=4577283

DanySK:
`DrawLayers` should be an interface, and the current `DrawLayers` should be an `AbstractDrawLayers` implementing it

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
paagamelo:
```java
// When I need to put something like this on multiple lines
if (condition1 && (condition2 || condition3)) {}
// I do something like this
if (condition1 && (condition2 
                   || condition3) {}
```
isn't it more clear that `condition3` is inside brackets?

If not, I will change it

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
paagamelo: It is indeed copy pasted from the `drawShape` class, I thought it was a good way to manage incarnations. In this particular case, inheritance cannot be used for obvious reasons, wouldn't it be jarring to use composition for such a small thing? 

One alternative is to get the current incarnation from the environment and remove this code completely, as well as the gui controls allowing the user to specify a particular incarnation

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/Utilities.java
paagamelo: `ConcreteIsolinesFactory` and `ConrecIsolinesFinder` use it. 

When receiving collections or arrays as input, I usually check whether or not they contain null elements. Is this too much?

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/conrec/Conrec.java
paagamelo: You mean `CASTAB`? I have no idea, I think they are using it to determine which `CASE` they are falling in, so the numbers stored in that array are basically different cases, but how the whole thing works is obscure. Should I shed light on it?

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
DanySK: No, it's horrible :)
Break on logic disjunction, preferably

```java
if (condition1
        && (condition2 || condition3) {}
```

if it gets too long, break using variables

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/Utilities.java
DanySK: Probably... It may make sense on a "very public" API, where it's likely to see misuse

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/conrec/Conrec.java
paagamelo: I looked for libraries with this particular algorithm, but didn't find anything. There are other libraries but with different algorithms not well explained. 
 
That code comes with a "copyright notice" on the top of the file (which seems to have been replaced by our copyright when I copy pasted the file on the project), in which authors grant permission to use the code as long as that notice is retained. Plus, they grant permission to modify the code "provided that such modifications are not distributed without the explicit consent of the authors". I actually have modified the code but for the purpose of making it respect our checkstyle and pmd rules. Plus, I don't think I am distributing those modifications, am I? No other license is released with the source code. The original comment:
```java
/*
 * Conrec.java
 *
 * Created on 5 August 2001, 15:03
 *
 * Copyright (c) 1996-1997 Nicholas Yue
 *
 * This software is copyrighted by Nicholas Yue. This code is base on the work of
 * Paul D. Bourke CONREC.F routine
 *
 * The authors hereby grant permission to use, copy, and distribute this
 * software and its documentation for any purpose, provided that existing
 * copyright notices are retained in all copies and that this notice is included
 * verbatim in any distributions. Additionally, the authors grant permission to
 * modify this software and its documentation for any purpose, provided that
 * such modifications are not distributed without the explicit consent of the
 * authors and that existing copyright notices are retained in all copies. Some
 * of the algorithms implemented by this software are patented, observe all
 * applicable patent law.
 *
 * IN NO EVENT SHALL THE AUTHORS OR DISTRIBUTORS BE LIABLE TO ANY PARTY FOR
 * DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES ARISING OUT
 * OF THE USE OF THIS SOFTWARE, ITS DOCUMENTATION, OR ANY DERIVATIVES THEREOF,
 * EVEN IF THE AUTHORS HAVE BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *
 * THE AUTHORS AND DISTRIBUTORS SPECIFICALLY DISCLAIM ANY WARRANTIES, INCLUDING,
 * BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
 * PARTICULAR PURPOSE, AND NON-INFRINGEMENT.  THIS SOFTWARE IS PROVIDED ON AN
 * "AS IS" BASIS, AND THE AUTHORS AND DISTRIBUTORS HAVE NO OBLIGATION TO PROVIDE
 * MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
 */
```

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/conrec/Conrec.java
DanySK: Obscure code inside something we maintain is super bad.
What we should do is the following:

1. check if a library doing what we need exists, if so, use it
2. otherwise, check the license of this source, if possible, make a library with it
3. use the library as a black box

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
DanySK: do the following:

1. incapsulate the behavior, so that both places use the same code
2. yes, change it so that the incarnation is deduced from the environment.

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/conrec/Conrec.java
DanySK: Yes you are. You are putting it on a public website, violating the license. This code must be isolated, incapsulated in a library, and then used.

paagamelo:
### About code duplication
Premise: drawing a layer's values (for instance, drawing a layer's isolines) only makes sense as long as this values are Numbers (or objects that can be mapped to numbers). Suppose the return type of `myLayer.getValue(Position)` is String, how could we draw `myLayer`? It makes no sense. Consequently, this mapping needs to be done somewhere. More generally, in order to draw a layer we need to map it to a `Function<? super P, ? extends Number>` where P is the position type.

The main issue with duplicated code is here:
DrawLayers
Both the highlighted classes contains the same code, which maps Bidimensional Gaussian Layers to functions described above. The solution is to incapsulate such behaviour in a dedicated object and then use it. So one solution would be the following:
Bridge
This is a sort of pattern Bridge. With this approach, it's easy to obtain the two effects we want:
```java
Effect drawBidiGausLayersIsolines = new DrawBidiGausLayersValues(new IsolinesDrawer())
Effect drawBidiGausLayersGradient = new DrawBidiGausLayersValues(new GradientDrawer())
```
The problem with this is that the gui needs a class with a parameterless constructor for each effect. In other words, we **can't** create our effects at run time, we need to create them at compile time. This is annoying, as it vanish the advantages of pattern Bridge, forcing us to define a class for each combination of `ValuesDrawer` and `DrawLayersValues`. Anyway, let's do it using a factory method:
Bridge2
This approach is lovely to me, but has a huge issue. We need to allow the user to specify how many isolines to draw, as well as how many samples to use when drawing gradient and other things too. So where are we supposed to put such controls? Ideally, we'd put them in the `IsolinesDrawer` and `GradientDrawer` classes, as they are directly related to them, and such classes need to access those values directly. However, by doing so, the controls wouldn't be exported in the gui, as (if I am correct) such annotation does not work through composition. An alternative is to put them in the `DrawBidiGausLayersIsolines` and `Gradient` classes. Nevertheless, this solution is jarring, as we are effectively vanishing the Bridge pattern advantages on our own.

How can we solve this issue? Let's think about this: on one side of the bridge there's the policy about how to map layers to functions, on the other side there's the policy about how to draw a function. One solution is to _invert_ the bridge, making the latter policy extend from the `DrawLayers` interface (or its abstract implementation, not showed here for simplicity), so it can specify the needed controls and they will be exported to the gui. So the UML would be the following:
Bridge3
This approach is less intuitive, but works better on our constraints. 

I am going with the second approach, but before I start coding I wanted to present this design issue to you. What do you think?

DanySK:
@paagamelo bravo, proposal approved.
Implementation side, *write in Kotlin whenever possible*. I'm tired of maintaining Java :)

Also, keep track of this stuff because it will be used in your thesis.

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
paagamelo: Ok, fixed

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
paagamelo: Ok, I removed all the empty lines (or added a comment). I will resolve that conversations too

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/isolines/Utilities.java
paagamelo: Ok, it's not used anymore

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayersIsolines.java
paagamelo: Fixed

alchemist/alchemist-swingui/src/main/java/it/unibo/alchemist/boundary/gui/effects/DrawLayers.java
paagamelo: The incarnation is now inferred from the environment in my code, I didn't incapsulate that behaviour in an object as it seemed overkill to me. We are talking about incapsulating a single line of code:
```java
env.getIncarnation().ifPresent(incarnation -> molecule = incarnation.createMolecule(molString));
```
If `@ExportForGUI` worked through composition, I could isolate not just that line of code, but all the related controls (such as `molString`, `molStringCached`, etc), which actually makes more sense

paagamelo:
Style fixed, duplication removed. As soon as the Conrec library will be on I will remove that code from the project and everything should be fine.

Also, the `ValuesDrawer` interface was renamed to `FunctionDrawer`. Regarding implementation, I wrote in Java so far just due to force of habit. I'm picking up Kotlin from now on. `LayerToFunctionMapper` and `BidiGausLayersMapper`, which I basically wrote from scratch, are in Kotlin

DanySK:
Is this ready for prime time?

paagamelo:
I was waiting for all the tests to pass, now it's ready

DanySK:
Nice job

PR #125, opened 2019-12-08
Fix SpecificPositions

### LINKED ISSUE #113: `SpecificPositions` bug
`SpecificPositions` is bugged as it calls `environment.makePosition()` with a single number.

Apparently, current impl

```
    private val positions: List<Position<*>> = positions.flatMap { it.map { environment.makePosition(it) } }
```
is not correct. A straightforward fix may be

```
    private val positions: List<Position<*>> = positions.map { environment.makePosition(*it.toList().toTypedArray()) }
```

fedpet:
Fixes #113

### DIFF
--- a/alchemist/alchemist-loading/src/main/kotlin/it/unibo/alchemist/loader/displacements/SpecificPositions.kt
+++ b/alchemist/alchemist-loading/src/main/kotlin/it/unibo/alchemist/loader/displacements/SpecificPositions.kt
@@ -10,12 +10,15 @@ package it.unibo.alchemist.loader.displacements
 import it.unibo.alchemist.model.interfaces.Environment
 import it.unibo.alchemist.model.interfaces.Position
 
-open class SpecificPositions(
+/**
+ * Given an environment and a list of list of numbers, it creates a list of the right position type for the environment.
+ */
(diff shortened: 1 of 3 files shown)


fedpet:
> Super nice. Could you also document it in the wiki?

It's already documented [https://alchemistsimulator.github.io/latest/wiki/usage/yaml/](here)

By the way I'm having some issues at understanding why the build is failing

DanySK:
Does it work on your system if you `./gradlew clean compileKotlin --parallel` ?

fedpet:
> Does it work on your system if you `./gradlew clean compileKotlin --parallel` ?

yes it does

DanySK:
Let's clean caches and re-run on CI, then.

DanySK:
The failure seems reproducible.

DanySK:
OK, there's actually a wrong test. Fixing.
(by the way, `compileTestKotlin` should fail on your machine as well)

fedpet:
since I ran `gradle clean`, it hasn't failed anymore on my machine

DanySK:
That's fixed, there's a ktlint rule ruining the party but it's overly pedantic and I'm suppressing it.

PR #126, opened 2019-12-08
Add ListOf variable

fedpet:
ListOf enables the following YAML syntax

```
variables:
  var:
    type: ListOf
    parameters: [1, 20, 30, 4000]
```


DanySK:
This seems to overlap with `ArbitraryVariable`, but  for the absence (implicitness?) of a default value... What am I missing?

fedpet:
> This seems to overlap with `ArbitraryVariable`, but for the absence (implicitness?) of a default value... What am I missing?

You are right, my bad. I must have overlooked it

PR #129, opened 2020-01-23
Add DEACCON algorithm for the generation of navigation meshes

paagamelo:
### The DEACCON algorithm
DEACCON (Decomposition of Environments for the Creation of Convex-region Navigation-meshes) is an algorithm capable of generating a navigation mesh of a given environment with obstacles. A navigation mesh is a collection of two-dimensional convex polygons representing which areas of an environment are traversable by agents (namely, walkable areas). Since convex polygons are generated, pedestrians can freely walk around within these areas, as it is guaranteed that no obstacle will be found. 
The deaccon algorithm is only able to deal with convex polygonal obstacles (concave ones can be be decomposed in convex meshes fairly easily, whereas for curves bounding boxes can be used). For a complete description of the algorithm see the [related paper](https://www.aaai.org/Papers/AIIDE/2008/AIIDE08-029.pdf). The algorithm implemented in Alchemist is a slight simplification of the one described in the linked paper, which assumes a rectangular shaped environment with euclidean geometry and double precision coordinates. Here's a brief description of how it operates:
- Seeding: a certain number of seeds is planted in the environment. Each seed is a square-shaped region that will grow maintaining a convex shape.

- Growing: planted seeds are extended until possible (i.e. until they are in contact with an obstacle or another seed on each side). The algorithm is capable to deal with non axis-aligned obstacles as well.

- Clean-up: adjacent regions are combined if the resulting polygon is still convex.

As the pictures show, performing only the above mentioned phases may result in a poor coverage of the walkable area. In order to avoid this, another operation called active seeding is performed. Basically, other seeds are generated, but much smaller:

Then the three phases described above are repeated for these newly generated seeds. Such seeds are called "active" because their purpose is to "cover" the holes resulting from the first three phases. Here's the final result:

### The code
The `geometry` package was enhanced quite a lot: the interfaces describing a convex polygon and a mutable convex polygon were added along with an implementation. A ton of useful functions was added to the `GeometryUtils.kt` file as well, which is now heavily used. The package `navigationmeshes.deaccon` was added inside the geometry one, containing the implementation of the deaccon algorithm. Such implementation depends a lot on the `ExtendableConvexPolygon` data structure, which is explicitly designed for the algorithm, and whose interface and implementation are contained in the same package. A part from that, a few tests for the new code were added, as well as a new effect for drawing navigation meshes in the gui. Here's a UML diagram describing the new structure of the geometry package:

deacconPR


DanySK:
Great documentation.
Can you amend the PR, adding documentation in the wiki as well?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
```diff
@@ -32,5 +36,15 @@ open class AvoidFlowField(
         this.map { it to layer.getValue(it).toDouble() }
             .filter { it.second < currentConcentration }
             .minBy { it.second }?.first ?: currentPosition
+    },
+    TargetSelectionStrategy {
+        with(env.getLayer(targetMolecule).get() as BidimensionalGaussianLayer) {
```
DanySK: What if there is no layer?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
```diff
@@ -32,5 +36,15 @@ open class AvoidFlowField(
...
+            val p = env.getPosition(pedestrian)
+            if (p == null) {
+                env.origin()
+            } else {
+                p + (p - Euclidean2DPosition(centerX, centerY))
```
DanySK: This breaks simulations using non-euclidean positions. Use env.makePosition instead

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
paagamelo: Now it returns the center of the environment

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
paagamelo: Changed to env.makePosition

paagamelo:
@DanySK I removed the PMD warning suppression and the useless initialisation. Obviously it didn't cause any NullPointerException, my fault. 
I added the documentation in the wiki as well, describing how the algorithm works and how to use it.

DanySK:
@paagamelo you can enable the CI on your own profile and trigger builds whenever you need without a build-triggering commit.

PR #135, opened 2020-02-22
Fix Simulation documentation

NiccoMlt:
As privately asked, I fixed `Simulation` interface Javadoc, specifically for `waitFor` method:

- Now doc clearly states that `waitFor` will return when timeout is over even if state is still different from expected;
- Now doc clearly states that synchronizing with `waitFor` does only synchronize with the internal simulation state, not with OutputMonitor update state;

Also:

- code was automatically formatted by Intellij IDEA;
- because of Intellij IDEA warning me to track personal configuration files that it auto-generated at project import, I updated the `.gitignore` documenting rules and ignoring those personal files.

### DIFF
--- a/alchemist/alchemist-interfaces/src/main/java/it/unibo/alchemist/core/interfaces/Simulation.java
+++ b/alchemist/alchemist-interfaces/src/main/java/it/unibo/alchemist/core/interfaces/Simulation.java
@@ -24,18 +24,15 @@ import it.unibo.alchemist.model.interfaces.Time;
  * This interface forces simulations to be independent threads, and make them
  * controllable from an external console.
  *
- * @param <T>
- *            Concentration type
- * @param <P>
- *            Position Type
+ * @param <T> Concentration type
+ * @param <P> Position Type
  */
 public interface Simulation<T, P extends Position<? extends P>> extends Runnable {
 
     /**
      * Adds an {@link OutputMonitor} to this simulation.
      *
-     * @param op
-     *            the OutputMonitor to add
+     * @param op the OutputMonitor to add
      */
     void addOutputMonitor(OutputMonitor<T, P> op);
 
@@ -90,17 +87,15 @@ public interface Simulation<T, P extends Position<? extends P>> extends Runnable
 
     /**
      * Executes a certain number of steps, then pauses it.
-     * 
-     * @param steps
-     *            the number of steps to execute
+     *
+     * @param steps the number of steps to execute
      */
     void goToStep(long steps);
 
     /**
      * Executes the simulation until the target time is reached, then pauses it.
-     * 
-     * @param t
-     *            the target time
+     *
+     * @param t the target time
      */
     void goToTime(Time t);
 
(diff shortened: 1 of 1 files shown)


alchemist/.gitignore
```diff
@@ -1,14 +1,43 @@
...
-gradle.xml
-modules.xml
-scala_compiler.xml
-vcs.xml
-compiler.xml
-misc.xml
-.idea/workspace.xml
-.idea/codeStyleSettings.xml
-.idea/inspectionProfiles/profiles_settings.xml
+*.ipr
```
DanySK: Why is this file in this PR?

NiccoMlt: Because, while importing Alchemist (following guidelines in README) in Intellij to improve the javadoc, git asked to track autogenerated settings files that should be ignored.
So I thought the first step should have been fix the gitignore to avoid this issue.

Should I open a separate pull request for the gitignore alone?

DanySK: Are the changes entirely safe? Where did you get them from? Do they change the inspection profile settings, or ignore them?

NiccoMlt: > Where did you get them from?

This is a merge I personally done between gitignore.io ignore configuration for JetBrains IDEs and the pre-existent .gitignore Alchemist used.
I habitually use [gitignore.io website](https://www.gitignore.io/) to generate `.gitignore` files and I never had a problem. The site uses patched versions of [GitHub gitignore repo](https://github.com/github/gitignore).
Also, JetBrains references the GitHub one in [official support](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems).

> Do they change the inspection profile settings, or ignore them?

With "inspection profile settings" you mean `.idea/inspectionProfiles/profiles_settings.xml`?
I merged it by ignoring `profiles_settings` XML files in any subfolder of the `.idea` configuration folder, as those files usually are user-dependent.
That specific file you mention was not tracked before, and it keeps being untracked.

> Are the changes entirely safe?

If there isn't anything weird in the project configuration, I think so.
In particular, files that now are tracked even if they should not will be:

- [.idea/codeStyles/codeStyleConfig.xml](https://github.com/AlchemistSimulator/Alchemist/blob/develop/alchemist/.idea/codeStyles/codeStyleConfig.xml)
- [.idea/copyright/profiles_settings.xml](https://github.com/AlchemistSimulator/Alchemist/blob/develop/alchemist/.idea/copyright/profiles_settings.xml)
- [.idea/dictionaries/danysk.xml](https://github.com/AlchemistSimulator/Alchemist/blob/develop/alchemist/.idea/dictionaries/danysk.xml)

Is that a problem in your opinion?

DanySK: Copyright and dictionary are an issue, yes. I usually check files one by one to make sure I know what's tracked/ignored. That's why I'm reluctant to merge these changes that untrack entire folders, I want some stuff to be there, like the Checkstyle / PMD / Spotbugs / Ktlint configuration, and the automated addition of copyright headers.

"Works for me" is not exactly a guarantee of quality I can rely upon for a large project such as Alchemist. At least comment line-by line what is being excluded, I want full control over that, and I might decide to track some configurations instead.

Maybe at this point a separate PR would be best, so that this one does not get stuck.

NiccoMlt: > "Works for me" is not exactly a guarantee of quality I can rely upon for a large project such as Alchemist.

I understand.

> Maybe at this point a separate PR would be best, so that this one does not get stuck.

Ok, I'll revert that specific commit in this PR

NiccoMlt: I rebased the branch dropping the undesired commit and force-pushed to my branch which I am doing the PR from.

If you think it might be useful, in the following days I will open a separate PR for the gitignore.
In that case, I think there might be some more improvements over the ones included in the rejected commit.

DanySK: Yes it makes sense

PR #161, opened 2020-03-31
Improve algorithm for the generation of navigation graphs

paagamelo:
The algorithm used to generate navigation graphs is now much more reliable and concise. Since it is only distantly related to the DEACCON algorithm, such name was abandoned in favour of a more meaningful one: NaviGator (Navigation Graphs generator).

NaviGator has two main differences compared to deaccon: it requires the user to specify the positions where to plant initial seeds (which means the problem of individuating different rooms is left to the user), and it is able to detect every axis-aligned crossing without further help from the user.

In order to obtain more accurate navigation graphs, the algorithm is now applied directly to images, so as to exploit integer coordinates (even though NaviGator can deal with double precision coordinates as well). The positions where to plant initial seeds can be specified directly in the image, marking each area with a given color. The image below shows the generation of the navigation graph of a simple building planimetry: the blue regions in the original image indicate where to plant initial seeds, these are then grown and crossings are found between them. The process is fully automatic.

navigation-graph-generation

Contextually, an `ImageEnvironment` capable of reading the visual format defined above was added. It is called `ImageEnvironmentWithGraph` and exposes a method allowing to obtain the generated graph. Here's a UML diagram related to the new class:

ImageEnvironmentWithGraph-UML

A part from what described above, orienting pedestrians and the associated behaviours were modified: they can now be placed only in an `Environment2DWithGraph`. A bunch of utils functions were added as well, and the wiki was modified accordingly to the new changes.


paagamelo:
Fixed detekt violations in `alchemist-implementationbase` and `alchemist-cognitive-agents`. Five violations have been suppressed:
- `MaxLineLength` violation in `RandomGeneratorUtils.kt`, because lines contain markdown links to webpages.
- `ArrayPrimitive` violation in `PositionUtils.kt`, this is about the usage of `Array<Primitive>` which leads to implicit boxing and performance hit. It has been suppressed because using a Kotlin DoubleArray (which is recommended), would have required its conversion to a primitive array anyway.
- `TooManyFunctions` violation in `GeometryUtils.kt`: what could be moved to a different file was moved, and in my opinion it makes no sense to split the remaining content. 
- `NestedBlockDepth` in `GraphUtils.kt`, in particular in the `primMinimumSpanningForest` function. This could be fixed quite easily, but would have caused more complexity in my opinion. Nested blocks are really easy to read in the prim's algorithm.
- `UndocumentedPublicProperty` violation in `AbstractCognitiveCarachteristic.kt`, a bunch of parameters (read from a configuration file) with self-explaining names were missing documentation. In my opinion, adding empty documentation or single line doc replicating or paraphrasing the name of the parameter is useless.

PR #266, opened 2020-04-11
Upgrade slf4j-api from 1.7.30 to 1.8.0-beta4

DanySK:
This update was prepared for you by UpGradle, at your service.


DanySK:
Upgrading SLF4J currently breaks logging.

PR #272, opened 2020-04-14
Re-enable HotSpot 14 under Windows

DanySK:


DanySK:
The build is affected by gradle/gradle#12681

PR #294, opened 2020-04-16
Divide steering actions into actions with a target and without

paagamelo:
Previously, the `SteeringAction` interface had a `target()` method which was meaningful only for some actions. This occasionally caused some unexpected behaviours and/or side effects which were incredibly time-expensive to debug.
Now, that interface has been divided into `SteeringAction` and `SteeringActionWithTarget`. Most of the steering actions were refactored due to that, and are now much cleaner. Here's a UML diagram of the top of the new hierarchy:

SteeringActionWithTargetUML

Additionally, the `Orienting` steering action was added. Previously, the orienting behaviour of a pedestrian was defined as a `Reaction`. However, representing it as an `Action` will be handy in the future, as it will allow us to treat the orienting behaviour as any other steering behaviour.
In spite of this, the implementation of the orienting behaviour as a reaction was maintained. The reason is that we're in a testing phase, and it's handy for me to have both implementation here, hence I'll allow some duplication which will be removed in the future.


DanySK:
Please also add notes to the website documentation :)

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractGroupSteeringAction.kt
```diff
@@ -0,0 +1,36 @@
...
+    protected open val env: Environment<T, P>,
+    reaction: Reaction<T>,
+    pedestrian: Pedestrian<T>
+) : GroupSteeringAction<T, P>, SteeringActionImpl<T, P>(env, reaction, pedestrian) {
+
+    /**
+     * Computes the centroid of the [group] in ABSOLUTE coordinates.
+     */
+    protected fun centroid(): P = with(group()) {
+        map { env.getPosition(it) }.reduce { acc, pos -> acc + pos } / size.toDouble()
```
DanySK, line 34: Uhmm... I see an operation here, this suggests `P` must be a vector. Do we need the bound to `AbstractEuclideanPosition` or can we work with `Vector<P>`?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+                else {
+                    val prevRoom = currRoom
+                    currRoom = newRoom
+                    val doorsLeadingBack = environment.graph().outgoingEdgesOf(currRoom)
+                        .filter { environment.graph().getEdgeTarget(it) == prevRoom }
+                    moveToClosestDoor(doors = doorsLeadingBack)
+                    return
+                }
+            }
+            else -> currRoom = newRoom
```
DanySK, line 192: `when` is functional and should not get used imperatively. Any reason not to use `if` here?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+        /*
+         * If a sub-destination of the route is in sight, updates the route removing
+         * all the sub-destination up to the one encountered.
+         */
+        if (route.isNotEmpty() && route.any { currRoom.contains(it.centroid) }) {
+            for (i in 0..route.indexOfLast { currRoom.contains(it.centroid) }) {
+                route.removeAt(0)
+            }
+        }
+        val rankings = if (route.isNotEmpty()) {
```
DanySK, line 211: route.takeUnless { it.isNotEmpty() }?.let { it.computeEdgeRankings(currRoom, route[0].centroid) }

looks cleaner IMHO

DanySK: and does not need an hybrid if with only one branch with curly brackets

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+            computeEdgeRankings(currRoom, route[0].centroid)
+        } else null
+        /*
+         * The pedestrian can see all the edges outgoing from the current room.
+         */
+        val minEdge = environment.graph().outgoingEdgesOf(currRoom)
+            .minWith(
+                compareBy({
+                    weight(it, rankings?.get(it))
```
DanySK, line 220: one liner

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+            targetDoor = minEdge
+            state = State.MOVING_TO_DOOR
+        }
+        /*
+         * Closed room, we can't move anywhere.
+         */
+        else {
+            state = State.ARRIVED
+        }
+    }
```
DanySK, line 240: Looks to me the goal of this method is to set the new state. As it is, it's very C-ish and not much kotlin-ish. A refactor with a `private fun nextState(...): State` computing the next state and this method reduced to a caller of `nextState` with possibly argument setup would be probably more readable.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+         * position of the pedestrian is always the same this will have no effect.
+         */
+        update()
+        return when (state) {
+            State.MOVING_TO_DOOR, State.CROSSING_DOOR, State.MOVING_TO_FINAL ->
+                interpolatePositions(currentPosition, subdestination)
+            /*
+             * The pedestrian doesn't want to move at all.
+             */
+            else -> currentPosition - currentPosition
```
DanySK, line 299: Ugh
if we don't have a "zero" or "origin" in `Vector`, then we must create it. We cannot subtract a vector from itself to get the origin, it's a kludge.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+            environment.graph().outgoingEdgesOf(area).distinct().count() <= 1
+
+    /*
+     * Picks the closest door among the provided ones (defaults to all doors)
+     * and sets all the state variables so as to move to the selected door.
+     * ifNoDoorCanBeFound is called if no door could be found, defaults to nothing.
+     */
+    private fun moveToClosestDoor(
+        doors: Collection<F> = environment.graph().edgeSet(),
+        ifNoDoorCanBeFound: () -> Unit = {}
```
DanySK, line 401: `noDoorAction`, maybe cleaner?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
```diff
@@ -0,0 +1,417 @@
...
+    ) {
+        val closestDoor = doors
+            .map { it to crossingPoint(it) }
+            .minBy { it.second.distanceTo(currentPosition) }
+        if (closestDoor != null) {
+            nextRoom = environment.graph().getEdgeTarget(closestDoor.first)
+            subdestination = closestDoor.second
+            targetDoor = closestDoor.first
+            state = State.MOVING_TO_DOOR
+        } else ifNoDoorCanBeFound.invoke()
```
DanySK, line 411: use curly brackets, and `noDoorAction()` looks better to me.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/ObstacleAvoidance.kt
```diff
@@ -2,50 +2,57 @@ package it.unibo.alchemist.model.implementations.actions
...
+            .map { obstacle: W ->
+                obstacle.nearestIntersection(currentPosition, target) to obstacle.bounds2D
+            }
+            .minBy { (intersection, _) -> currentPosition.distanceTo(intersection) }
+            ?.let { (intersection, bound) -> intersection to env.makePosition(bound.centerX, bound.centerY) }
+            ?.let { (intersection, center) -> (intersection - center).resizedToMaxWalkIfGreater() }
+            /*
+             * Otherwise we just don't apply any repulsion force.
+             */
+            ?: env.makePosition(0.0, 0.0)
```
DanySK, line 46: Vector.origin

paagamelo:
> Please also add notes to the website documentation :)

Concerning what? Nothing changed regarding the writing of simulations with cognitive pedestrians.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Orienting.kt
```diff
@@ -0,0 +1,179 @@
...
+    /*
+     * A rough estimation of the area of a shape.
+     */
+    private fun Shape.area(): Double =
+        with(bounds2D) {
+            abs(width * height)
+        }
+
+    companion object {
+        private const val wallRepulsionFactor = 0.3
```
DanySK, line 177: How was this computed?
We need this to be public, documented, and modifiable. Otherwise it's a hidden parameter

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/SteeringActionImpl.kt
```diff
@@ -1,63 +1,60 @@
...
-    env,
-    pedestrian,
-    routing,
-    target,
-    speed
-), SteeringAction<T, P> {
-
-    override fun cloneAction(n: Node<T>, r: Reaction<T>) =
-        SteeringActionImpl(env, r, n as Pedestrian<T>, target, speed, routing)
+abstract class SteeringActionImpl<T, P>(
```
DanySK, line 22: Refactor to `AbstractSteeringAction`. Avoid `Impl` when possible, and make non-instantiability clear from name.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/SteeringActionImpl.kt
```diff
@@ -1,63 +1,60 @@
...
+     * Keeping bounds as much general as we can (instead of using AbstractEuclideanPosition),
+     * so as to allow [AbstractOrienting] to derive from this class.
+     */
+    where
+        P : Position<P>,
+        P : Vector<P> {
 
     /**
-     * Next RELATIVE position.
+     * Next relative position.
```
DanySK, line 42: No need to scream, indeed.
Thanks :)

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/steeringstrategies/Weighted.kt
```diff
@@ -31,9 +32,17 @@ open class Weighted<T>(
...
+     * and we pick the action whose target is closest to the current position of the pedestrian,
+     * which will be considered the overall target.
+     */
     override fun computeTarget(actions: List<SteeringAction<T, Euclidean2DPosition>>): Euclidean2DPosition =
         with(env.getPosition(pedestrian) ?: env.origin()) {
-            actions.map { it.target() }.minBy { it.distanceTo(this) } ?: this
+            actions
+                .filterIsInstance<SteeringActionWithTarget<T, out Euclidean2DPosition>>()
+                .map { it.target() }
+                .minBy { it.distanceTo(this) } ?: this
```
DanySK, line 45: ?: this on a new line.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/AbstractOrientingBehavior.kt
```diff
@@ -202,9 +211,7 @@ abstract class AbstractOrientingBehavior<T, P, A, N, E, M, F>(
         }
         val rankings = if (route.isNotEmpty()) {
             computeEdgeRankings(currRoom, route[0].centroid)
-        } else {
-            null
-        }
+        } else null
```
DanySK, line 214: I feel like I've seen this code before. There is duplication: can you extract this "rankings" into a protected method in a common abstract class?

DanySK: (or a protected extension function of Route, whatever is cleaner)

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/AbstractOrientingBehavior.kt
```diff
@@ -364,6 +371,24 @@ abstract class AbstractOrientingBehavior<T, P, A, N, E, M, F>(
...
+        val currentPosition = environment.getPosition(pedestrian)
+        val closestDoor = doors
+            .map { it to computeSubdestination(it) }
+            .minBy { it.second.distanceTo(currentPosition) }
+        if (closestDoor != null) {
+            nextRoom = environment.graph().getEdgeTarget(closestDoor.first)
+            subdestination = closestDoor.second
+            targetDoor = closestDoor.first
+            state = State.MOVING_TO_DOOR
+        } else ifNull.invoke()
```
DanySK, line 389: This also looks very similar to another snippet of code I already commented, the one with `ifNoDoorIsFound`. Duplication is a source for bugs. Please check if you can extract the behaviour in a common superclass. If you can't, we'll use strategy.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FlowFieldSteeringAction.kt
```diff
@@ -2,45 +2,68 @@ package it.unibo.alchemist.model.implementations.actions
...
+     */
+    abstract fun List<Euclidean2DPosition>.selectPosition(
+        layer: Layer<Number, Euclidean2DPosition>,
+        currentConcentration: Double
+    ): Euclidean2DPosition
+
+    /**
+     * Discard the positions the node can't fit.
+     */
+    protected fun MutableList<Euclidean2DPosition>.discardUnsuitablePositions(): List<Euclidean2DPosition> =
```
DanySK, line 56: Why "Mutable"? You are not mutating anything. This works on lists.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FlowFieldSteeringAction.kt
```diff
@@ -2,45 +2,68 @@ package it.unibo.alchemist.model.implementations.actions
...
+        map {
+            if (env is EuclideanPhysics2DEnvironmentWithObstacles<*, Number>) {
```
DanySK, line 58: Uhmm... don't we have a superinterface for environments with obstacles? we just need `next`
An the if may become a one-liner

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Flee.kt
```diff
@@ -10,18 +12,28 @@ import it.unibo.alchemist.model.interfaces.Reaction
...
  * @param env
  *          the environment inside which the pedestrian moves.
+ * @param reaction
+ *          the reaction which executes this action.
  * @param pedestrian
  *          the owner of this action.
  * @param coords
  *          the coordinates of the position the pedestrian moves away.
  */
 open class Flee<T, P : AbstractEuclideanPosition<P>>(
```
DanySK, line 22: To we need this specificity? Can't we set a bound on `Vector<P`> and `Position<P>`?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Combine.kt
```diff
@@ -21,21 +22,18 @@ import it.unibo.alchemist.model.interfaces.movestrategies.TargetSelectionStrateg
  *          the logic according to the steering actions are combined.
  */
 class Combine<T, P : AbstractEuclideanPosition<P>>(
```
DanySK, line 24: Double bound on `Position<P>` and `Vector<P>`, doable?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractGroupSteeringAction.kt
paagamelo: `P` is a `Position` as well, I could bound `P` to `Vector` and `Position` to maintain more generality. Even though pragmatically this means using `AbstractEuclideanPosition`.

DanySK:
> > Please also add notes to the website documentation :)
> 
> Concerning what? Nothing changed regarding the writing of simulations with cognitive pedestrians.

"the Orienting steering action was added"

No impact at all?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractGroupSteeringAction.kt
DanySK: Yes, but I prefer to use interfaces whenever possible. Allows for more flexibility in case of refactorings.

paagamelo:
> "the Orienting steering action was added"
> 
> No impact at all?

No, as we maintained the implementation of the orienting behaviour as a `Reaction`, which is the one used in the wiki and in our tests as well. In the future, when we'll have an effective way to mix steering actions, my plan is to remove the `OrientingBehavior` reaction, modify our tests and the wiki accordingly.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractOrienting.kt
paagamelo: The goal of `inNewRoom()` is to set the new state and the related variables (such as `subdestination`, `currRoom`, etc) when the agent is in a new room. It's part of the finite state machine describing the orienting behaviour. I agree it's very C-ish, but I doubt whether defining a `private fun nextState(...): State` would make it any cleaner, as that method would be almost a copy-paste of the current one. 
I propose to maintain the policy regarding the setting of state variables within this method, and export anything else in small new methods, as for example `updateRoute()` and `computeRankings()`. It won't make the method more Kotlin-ish, but probably more readable.

DanySK: ok

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Orienting.kt
paagamelo: It was found empirically, I moved it to the constructor and documented it.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/AbstractOrientingBehavior.kt
paagamelo: Yes there's a lot of duplication, there are two "versions" of the orienting behaviour: one deriving from `Action` and the other deriving from `Reaction`. I plan to remove the latter, but only after finding an effective method for mixing steering actions. Given that, I can maintain the duplication until that moment, or I can extract the common code in a superclass. Take into account that such superclass will be deleted in the future with every probability, so it will take some time to create it and to delete it as well. If you think it's best to collect the common code anyway, I'll define a superclass.

DanySK: Uhmm... I get it. Can't we factor it out into a strategy?
I'd like duplication issues to be solved before getting this into the mainline. We could wait for the steering mixing to be in place, though, to avoid doing things twice.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractGroupSteeringAction.kt
paagamelo: True, I relaxed the bounds where possible.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FlowFieldSteeringAction.kt
paagamelo: With "one-liner" you mean `if .. else ..` on a single line or `map { if ..`?

DanySK: possibly the latter
If the type name is short, you can
`map { if (env is X<A, B>) somemethod() else null }`
going to a single new line in a `map` rarely works, you gain a single column due to indentation.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/AbstractOrientingBehavior.kt
paagamelo: I isolated the common code in a new class which is treated more or less like a strategy.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FlowFieldSteeringAction.kt
paagamelo: Ok, I moved the if statement and its condition on the same line of `map`. Branches are too long to be contained in a single line.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
```diff
@@ -365,6 +401,6 @@ abstract class AbstractOrientingBehavior<T, P, A, N, E, M, F>(
             environment.graph().outgoingEdgesOf(area).distinct().count() <= 1
 
     companion object {
-        private const val destinationWeight = 0.1
+        private const val finalDestinationWeight = 0.1
```
DanySK, line 404: Why 0.1? Looks like this is an hidden parameter...

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
```diff
@@ -144,86 +159,93 @@ abstract class AbstractOrientingBehavior<T, P, A, N, E, M, F>(
...
+     * noDoorAction is called if no door could be found, defaults to nothing.
+     */
+    private fun moveToClosestDoor(
+        doors: Collection<F> = environment.graph().edgeSet(),
+        noDoorAction: () -> Unit = {}
+    ): Unit = doors
+        .map { it to crossingPoint(it) }
+        .minBy { it.second.distanceTo(currentPosition) }
+        ?.let { (closestDoor, crossingPoint) -> moveToDoor(closestDoor, crossingPoint) }
+        ?: noDoorAction.invoke()
```
DanySK, line 192: do you need invoke? Can't you just `noDoorAction()`?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
```diff
@@ -1,51 +1,56 @@
...
+) : FlowFieldSteeringAction<P>(env, reaction, pedestrian, targetMolecule)
+    where
+        P : Position2D<P>,
+        P : Vector2D<P> {
+
+    override fun cloneAction(n: Node<Number>, r: Reaction<Number>): Action<Number> =
+        AvoidFlowField(env, r, n as Pedestrian2D<Number>, targetMolecule)
+
+    override fun List<P>.selectPosition(layer: Layer<Number, P>, currentConcentration: Double): P = this
+        .toMutableList()
```
DanySK, line 40: Considering that you modify the collection multiple times, have you considered using `Sequence` instead?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FollowFlowField.kt
```diff
@@ -1,50 +1,56 @@
...
-                }
+) : FlowFieldSteeringAction<P>(env, reaction, pedestrian, targetMolecule)
+    where
+        P : Position2D<P>,
+        P : Vector2D<P> {
+
+    override fun cloneAction(n: Node<Number>, r: Reaction<Number>): Action<Number> =
+        FollowFlowField(env, r, n as Pedestrian2D<Number>, targetMolecule)
+
+    override fun List<P>.selectPosition(layer: Layer<Number, P>, currentConcentration: Double): P = this
```
DanySK, line 39: This method is duplicated... can its logic get refactored out?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
paagamelo: I removed it, so we have fewer parameters.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
paagamelo: I can! Fixed.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
paagamelo: Moved to `Sequence`, I didn't remember they were faster.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FollowFlowField.kt
paagamelo: It's not a duplicate. `AvoidFlowField.selectPosition` selects the position with the least concentration, whereas `FollowFlowField.selectPosition` picks the highest one. A super method taking in a boolean would need to have at least two `if/else` and I think it wouldn't be so readable.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
DanySK: So this is no longer a parameter? Was it unused? Where have this 0.1 gone?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FollowFlowField.kt
DanySK: not really.
This portion looks completely identical to me
```
        .apply {
            if (layer is BidimensionalGaussianLayer<*>) {
                /*
                 * If the layer is Gaussian (i.e. has a center), probably the most suitable
                 * position is the one obtained by moving away from the center along the
                 * direction which connects the current position to the center.
                 */
                val center = env.makePosition(layer.centerX, layer.centerY)
                this.add(currentPosition + (currentPosition - center).resized(maxWalk()))
            }
        }
        .discardUnsuitablePositions(env, pedestrian)
        .map { it to layer.concentrationIn(it) }
```

```
        .apply {
            if (layer is BidimensionalGaussianLayer<*>) {
                /*
                 * If the layer is Gaussian (i.e. has a center), probably the most suitable
                 * position is the one obtained by moving away from the center along the
                 * direction which connects the current position to the center.
                 */
                val center = env.makePosition(layer.centerX, layer.centerY)
                this.add(currentPosition + (center - currentPosition).resized(maxWalk()))
            }
        }
        .discardUnsuitablePositions(env, pedestrian)
        .map { it to layer.concentrationIn(it) }
```
there is no need to pass booleans, it just needs to get factored out.

paagamelo: There's a difference, here we add a position moving away from the center as the crow flies:

`this.add(currentPosition + (currentPosition - center).resized(maxWalk()))`

while here we add a position moving towards the center:

`this.add(currentPosition + (center - currentPosition).resized(maxWalk()))`

Maybe I could factor out the code computing the center of the layer.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
paagamelo: That parameter came from Andresen's paper and it was used to lower the weight of edges leading to final destinations. In brief: when weighting edge, the final destination factor assumed value `0.1` if the edge leaded to a final destination, `1.0` otherwise (since factors are multiplied, `1.0` is the identity-value or something - can't remember how it's called). This allowed pedestrians to consider (and take) exits they eventually found along the way. The idea behind `0.1` (at least this is what I suppose, as Andresen didn't documented that so much) is that it should lower the weight of an exit so as to force a pedestrian to choose that edge, but without voiding the weight (i.e. it couldn't simply be `0`), so as to allow different weights based on other factors in case of multiple exits.
The final destination factor was a good way of taking into account exits found along the way. However, I removed it. Now our pedestrians explicitly look for exits they may found along the way (in particular, whenever they're in a new room). If one or more exits are found, they are weighted with a _different_ weighting function, and one of them is taken. The weighting function for exits at present takes into account the distance of an exit and the congestion. This allows us to consider different factors when weighting an exit, which is good because e.g. we don't care about the cognitive map factor, which was previously taken into account as well.
What do you think?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FollowFlowField.kt
DanySK: yes please

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
DanySK: If the parameter is gone, then it's ok. From the diff it looks to me it's just renamed, though.
If this parameter still circulates (but this is general for every parameter), then:
* reify it as public in a companion object
* add it as parameter in the @jvmOverloads constructor, defaulting to the default value of the companion
* document the default value citing the source of the default value, so that the value origin can be traced, and does not look like something magic that we threw in to make the machinery work.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/FollowFlowField.kt
paagamelo: Ok, did it.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/AbstractOrientingBehavior.kt
paagamelo: Ok, in the future I'll do that.
Concerning our `0.1`, it's not renamed but completely removed. If you want, point me the code that made you think so.

PR #368, opened 2020-05-13
Reorganize orienting behaviour 

paagamelo:
Multiple changes have been made.

### Reorganize orienting behaviour
`OrientingBehavior` class was replaced by multiple classes each one containing a well defined and documented portion of the original logic. Two fundamental interfaces were added:
- `NavigationAction` describes a steering action allowing to navigate the environment consciously (e.g. without getting stuck in U-shaped obstacles),
- `NavigationStrategy` defines what to do when entering a new room.

A base `NavigationAction` was implemented, this takes care of _how_ to properly move the pedestrian in the environment while delegating the decision on _where_ to move it to a `NavigationStrategy`, as shown in the UML diagram below.

new-orienting-behaviors1

Different strategies implement different behaviours, and complex behaviours are obtained reusing simpler ones. Here's an overview of the main strategies available.

new-orienting-behaviors2

These changes aim to allow a wide reuse of these behaviours.

### Add `SinglePrevalent` steering strategy
A new `SteeringStrategy` was added (those strategies define how to combine multiple `SteeringAction`s): `SinglePrevalent` accepts a list of actions where one is considered to be prevalent - it has to be a `NavigationAction`, as they guarantee to navigate the environment consciously. The purpose of the new strategy is to combine the potentially contrasting forces to which the pedestrian is subject while maintaining that warranty, how this is achieved is widely explained in the api documentation.

### Refactor `ConvexPolygon` hierarchy
API documentation was improved and an abstract class was added on top of the hierarchy.

### Replace `DoubleInterval` with Kotlin's `Range` extension functions
This allows to use idiomatic expressions such as `in` and `..`, but spreads the API to multiple extension functions.. I'm not sure it was the best thing to do.

### Others
The wiki was modified according to the new changes, some tests were modified and others were added.


DanySK:
I'll try to review this next week. In the meantime, please, take care of issues spotted by [codacy](https://app.codacy.com/manual/Alchemist/Alchemist/pullRequest?bid=18083680&prid=5519299&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&) and [codefactor](https://www.codefactor.io/repository/github/alchemistsimulator/alchemist/pull/368)

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/IterableExtension.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Takes the first [percentage] * size elements of the list.
+ */
+fun <T> List<T>.takePercentage(percentage: Double): List<T> = take((percentage * size).toInt())
+
+/**
+ * Converts an array of numbers representing [Euclidean2DPosition]s to an actual list of positions.
+ */
+fun Array<out Number>.toPositions(): List<Euclidean2DPosition> =
+    toList().chunked(2) { Euclidean2DPosition(it[0].toDouble(), it[1].toDouble()) }
```
DanySK, line 23: Woah, this is broken under many conditions. How frequently do you use it? If you use it in a single place, move it there.
Otherwise, you need to pass the environment as parameter, chunk by dimensions and use makePosition to build the correct position.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Seek2D.kt
```diff
@@ -28,7 +29,7 @@ open class Seek2D<T, P>(
         return (current.surrounding(env, maxWalk) + superPosition)
             .asSequence()
             .discardUnsuitablePositions(env, pedestrian)
-            .minBy { it.distanceTo(super.target()) }?.minus(current)
-            ?: currentPosition
+            .minBy { it.distanceTo(target) }?.minus(current)
+            ?: env.origin()
```
DanySK, line 33: why?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Utils.kt
```diff
@@ -37,3 +40,27 @@ fun <T, P> Sequence<P>.discardUnsuitablePositions(
          */
         environment !is PhysicsEnvironment<T, P, *, *> || environment.canNodeFitPosition(pedestrian, it)
     }
+
+/**
+ * A delegate allowing to lazily initialise a mutable variable (= var).
+ */
+class LazyMutable<T>(private val initializer: () -> T) : ReadWriteProperty<Any?, T> {
```
DanySK, line 47: Reify this in its own file. I do not like having "Utils" files stuffed with a lot of stuff. And by the way this is so generic that I wonder why it's not in library... is there some side effect to its existence and use?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/Exploring.kt
```diff
@@ -0,0 +1,126 @@
...
+    override val action: EuclideanNavigationAction<T, N, E, ConvexPolygon, Euclidean2DPassage>,
+    /**
+     * Weight assigned to known impasses, see [impasseFactor].
+     */
+    private val impasseWeight: Double = DEFAULT_IMPASSE_WEIGHT
+) : EuclideanNavigationStrategy<T, N, E, ConvexPolygon, Euclidean2DPassage> {
+
+    companion object {
+        /**
+         * Empirically found to work well (allows the pedestrian to avoid known impasses).
```
DanySK, line 47: Please add a note on the wiki as well.
Maybe a tabular with all the hardcoded parameters and their meaning? I do not want them to slip away when we write the paper

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/KnownDestinationReaching.kt
```diff
@@ -0,0 +1,140 @@
...
+         * In order, we try to find a path in which:
+         * - the first landmark is in current room and the last one in destination room
+         * - the first landmark is in current room and the last one in a room adjacent to destination room
+         * - the first landmark is in a room adjacent to current room and the last one in destination room
+         * - the first landmark is in a room adjacent to current room and the last one in a room adjacent to destination
+         * room
+         * - landmarks may be located anywhere (and may coincide), but we try to minimise the distance between the first
+         * landmark and currPos and the distance between the last landmark and destination
+         */
+        return (cartesianProduct(landmarksInCurrRoom, landmarksInDestRoom) +
```
DanySK, line 121: can you factor out this complicate return into its own function? Possibly two functions, one for building the sequence, and one for processing it

alchemist/alchemist-cognitive-agents/build.gradle.kts
```diff
@@ -12,6 +12,7 @@ dependencies {
     implementation(project(":alchemist-influence-sphere"))
     implementation(Libs.konf)
     implementation(Libs.jgrapht_core)
+    implementation(Libs.boilerplate)
```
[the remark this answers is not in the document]
paagamelo: It was used in a "middle version" of the navigation framework, but now is unused: I removed it.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/cognitiveagents/characteristics/cognitive/OrientingAgent.kt
```diff
@@ -19,23 +28,41 @@ interface OrientingAgent<V, A, N, E> where
...
+     */
+    fun <M : ConvexGeometricShape<V, A>> registerVisit(area: M) {
+        volatileMemory[area] = (volatileMemory[area] ?: 0) + 1
+    }
+
+    /**
+     * Unregisters a visit to the provided [area] in the agent's [volatileMemory].
+     */
+    fun <M : ConvexGeometricShape<V, A>> unregisterVisit(area: M) {
+        volatileMemory[area]?.let { volatileMemory[area] = it - 1 }
```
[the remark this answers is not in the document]
paagamelo: `volatileMemory[area]?.minus(1)`. Anyway, similarly to the boilerplate dependency this is not used anymore: I'll remove it.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractNavigationAction.kt
```diff
@@ -0,0 +1,250 @@
...
+     * shape. In the future, something better could be done.
+     */
+    protected val minDistance: Double = pedestrian.shape.diameter
+
+    /**
+     * @returns true if the distance to [pedestrianPosition] is smaller than or equal to [minDistance].
+     */
+    protected open fun P.isReached(): Boolean = distanceTo(pedestrianPosition) <= minDistance
+
+    protected enum class State {
```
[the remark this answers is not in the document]
paagamelo: Does `NavigationState` sounds good?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractNavigationAction.kt
```diff
@@ -0,0 +1,250 @@
...
+     * Updates [pedestrianPosition] and [currentRoom], this can be costly.
+     * Depending on how [ConvexGeometricShape.contains] manage points on the boundary, the pedestrian could
+     * be inside two (adjacent) rooms at once. This can happen in two cases:
+     * - when in [State.MOVING_TO_CROSSING_POINT_1] or [State.MOVING_TO_FINAL] and the pedestrian is moving
+     * on [previousRoom]'s boundary. In such case [previousRoom] is used.
+     * - when crossing a door or in [State.NEW_ROOM] and [expectedNewRoom] is adjacent to [previousRoom].
+     * In such case [expectedNewRoom] is used.
+     * Otherwise the first room containing [pedestrianPosition] is used.
+     */
+    protected open fun updateCachedVariables() {
```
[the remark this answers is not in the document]
paagamelo: It's unit

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
```diff
@@ -21,12 +25,15 @@ import it.unibo.alchemist.model.interfaces.geometry.Vector2D
  *          the owner of this action.
  * @param targetMolecule
  *          the {@link Molecule} you want to know the concentration in the different positions of the environment.
+ * @param viewDepth
+ *          the depth of view of the pedestrian, defaults to infinity.
  */
-open class AvoidFlowField<P>(
+class AvoidFlowField<P> @JvmOverloads constructor(
```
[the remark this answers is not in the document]
paagamelo: I have no idea, added `open` back.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractNavigationAction.kt
```diff
@@ -0,0 +1,250 @@
...
+        pedestrianPosition = environment.getPosition(pedestrian)
+        currentRoom = when {
+            (state == State.MOVING_TO_CROSSING_POINT_1 || state == State.MOVING_TO_FINAL) &&
+                previousRoom.orFail().contains(pedestrianPosition) -> previousRoom
+            (state == State.MOVING_TO_CROSSING_POINT_2 || state == State.CROSSING_DOOR || state == State.NEW_ROOM) &&
+                expectedNewRoom.isPresent &&
+                expectedNewRoom.get().contains(pedestrianPosition) -> expectedNewRoom
```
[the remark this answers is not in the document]
paagamelo: It is not, but I though it was better to use optionals in general. Do they have any performance hit?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
```diff
@@ -35,6 +42,15 @@ open class AvoidFlowField<P>(
     override fun cloneAction(n: Node<Number>, r: Reaction<Number>): Action<Number> =
         AvoidFlowField(env, r, n as Pedestrian2D<Number>, targetMolecule)
 
+    /**
+     * Moves the pedestrian only if he percepts the danger (either because it is in sight or
+     * due to social contagion).
+     */
+    override fun nextPosition(): P = when {
+        pedestrian.wantsToEvacuate() || isDangerInSight() -> super.nextPosition()
+        else -> env.origin()
```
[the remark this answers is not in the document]
paagamelo: It does not: `nextPosition()` returns the next _relative_ position, `env.origin()` is the zero vector, meaning the pedestrian won't move.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/IterableExtension.kt
paagamelo: It's used widely, fixed as indicated.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Seek2D.kt
paagamelo: The next _relative_ position has to be returned: if we don't want to move, `env.origin()` (= the zero vector) is what we need. Returning `currentPosition` was an error.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/BaseEuclideanNavigationAction.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    @Suppress("UNCHECKED_CAST") // as? operator is safe
+    override fun cloneAction(n: Node<T>?, r: Reaction<T>?): Action<T> {
+        require(n as? OrientingPedestrian<T, Euclidean2DPosition, Euclidean2DTransformation, N, E> != null) {
+            "node not compatible, required: " + pedestrian.javaClass + ", found: " + n?.javaClass
+        }
+        n as OrientingPedestrian<T, Euclidean2DPosition, Euclidean2DTransformation, N, E>
+        require(r != null) { "reaction can't be null" }
+        BaseEuclideanNavigationAction(environment, r, n, wallRepulsionFactor).let { copy ->
+            copy.strategy = this.strategy
```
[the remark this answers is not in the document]
paagamelo: It is mutable.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/Utils.kt
paagamelo: Moved to its own file. There are no side effects as far as I know, probably it's not in the library because you can have a `lateinit var`, even though `lateinit` doesn't allow you to specify a lazy initialiser, which is why I created that class.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/KnownDestinationReaching.kt
paagamelo: Created two functions (more or less) as indicated.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/Exploring.kt
paagamelo: I added two tables: one describing the factors used by the weighting system, another containing all the hardcoded parameters.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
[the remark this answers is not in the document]
paagamelo: I'll add a comment on that.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractNavigationAction.kt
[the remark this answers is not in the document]
DanySK: Yeah much better... but still generic, it could be a car. I see that's protected however, so just NavigationState is ok

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractNavigationAction.kt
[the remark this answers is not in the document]
DanySK: Performance impact of optionals are irrelevant in this context.
Yes it's much better to use optionals over nullability in Java or Scala, since `null` eludes the type system (and by the way makes the entire type system unsound if used in conjunction with generics, see [this paper](https://dl.acm.org/doi/10.1145/2983990.2984004)). Kotlin takes a different stance (inspired by Ceylon) and provides nullable types. It's very much idiomatic to use them.

As a rule of thumb: if the consumer of your code may well be Java or Scala, prefer Optional. If your consumer is meant to be Kotlin (only or largely) then favor nullable types.

DanySK:
also, keep in sync with develop

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AvoidFlowField.kt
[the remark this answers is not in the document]
DanySK: no wait, I'm not referring to `open`, I'm referring to `@JvmOverloads`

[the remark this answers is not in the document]
paagamelo: I added `@JvmOverloads`, not removed. It's needed when the constructor has default parameters.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractNavigationAction.kt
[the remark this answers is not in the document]
paagamelo: Perfect, thanks for the explanation. I'll replace optionals with nullables.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/ReachDestination.kt
```diff
@@ -0,0 +1,71 @@
...
+    reaction: Reaction<T>,
+    pedestrian: OrientingPedestrian<T, Euclidean2DPosition, Euclidean2DTransformation, N, E>,
+    vararg destinations: Number
+) : BaseEuclideanNavigationAction<T, N, E>(environment, reaction, pedestrian) {
+
+    override var strategy: EuclideanNavigationStrategy<T, N, E, ConvexPolygon, Euclidean2DPassage> by lazyMutable {
+        destinations
+            .toPositions(environment)
+            .partition { inferIsKnown(it, pedestrian, environment) }
+            .let { (known, unknown) ->
```
DanySK, line 51: one liner

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/ReachDestination.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+    }
+}
+
+/**
+ * Infers if a [destination] is known by the [pedestrian] (see [Pursuing]). A destination is considered
+ * to be known if the pedestrian's cognitive map contains at least one landmark located in the same
+ * room (= [environment]'s area) of the destination, or in an adjacent room.
+ */
+fun <T, P, N : ConvexGeometricShape<P, *>, M : ConvexGeometricShape<P, *>> inferIsKnown(
```
DanySK, line 62: that's a strange place for a function. Where is it used?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/DynamicPursuing.kt
```diff
@@ -0,0 +1,50 @@
...
+        destination = newDestination
+        action.currentRoom?.let {
+            /*
+             * If the pedestrian is inside a room, we force the re-computation of what to do. Otherwise
+             * he/she's crossing a door and inNewRoom will be called as soon as a room is reached.
+             */
+            inNewRoom(it)
+        }
+        if (voidVolatileMemory) {
+            pedestrian.volatileMemory.replaceAll { _, _ -> 0 }
```
DanySK, line 47: why not just `clear`?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/KnownDestinationReaching.kt
```diff
@@ -0,0 +1,148 @@
...
+ * landmark and the pedestrian's position and the distance between the last landmark and the destination. In
+ * spite of this, the first landmark may be far from the pedestrian's position and the same applies to the
+ * last landmark and the destination. This means the pedestrian may not have information regarding any path
+ * leading from its current position to the first landmark and from the last one to the destination (such a
+ * path may not exist at all).
+ * If the cognitive map is poor the returned path may be poor. On average when the first landmark is farther
+ * from the pedestrian's position than the destination itself it's just more convenient to pursue the latter.
+ * If the cognitive map is empty the path will be empty as well.
+ */
+fun <T, P, N : ConvexGeometricShape<P, *>, M : ConvexGeometricShape<P, *>> findKnownPath(
```
DanySK, line 91: put this one and the followign either inside the `companion`, or (if access to members is useful to reduce parameterization) as a function of the class. Protect the visibility if it is not used elsewhere (`private`).
On a second thought, definitely inside the class, to reuse all the generics.

`findKnownPath` can be an extension method of P

DanySK: and maybe at this point could be something like `P.knownPathTo`

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/KnownDestinationReaching.kt
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * @returns a sequence of pairs of landmarks, for each pair the first element is contained in [startRoom] or
+ * in an adjacent one, and the second element is contained in [endRoom] or in adjacent one. The sequence is
+ * sorted, pairs appear in the following order:
+ * - pairs where the first landmark is inside [startRoom] and the second landmark is inside [endRoom],
+ * - pairs where the first landmark is inside [startRoom] and the second landmark is adjacent to [endRoom],
+ * - pairs where the first landmark is adjacent to [startRoom] and the second landmark is inside [endRoom],
+ * - pairs where the first landmark is adjacent to [startRoom] and the second landmark is adjacent to [endRoom].
+ */
+fun <T, P, N : ConvexGeometricShape<P, *>, M : ConvexGeometricShape<P, *>> buildSequence(
```
DanySK, line 130: same as above

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/NavigationPrioritisedSteering.kt
```diff
@@ -0,0 +1,55 @@
...
+     */
+    alpha: Double = SinglePrevalent.DEFAULT_ALPHA
+) : SteeringBehavior<T>(
+    env,
+    pedestrian,
+    timeDistribution,
+    SinglePrevalent(
+        env,
+        pedestrian,
+        {
```
DanySK, line 45: Hard to read: reify in a named function (in the companion) and use a method reference

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/utils/LazyMutable.kt
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Optional
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+/**
+ * A delegate allowing to lazily initialise a mutable variable (= var).
+ */
+class LazyMutable<T>(private val initializer: () -> T) : ReadWriteProperty<Any?, T> {
+
+    private var value: Optional<T> = Optional.empty()
```
DanySK, line 21: Prefer `T?`, that's entirely internal.
Plus, this one is a candidate for being exported into a separate library. Can you open an issue about, once the PR gets merged?

alchemist/alchemist-cognitive-agents/src/test/kotlin/it/unibo/alchemist/test/TestOrientingBehavior.kt
```diff
@@ -2,50 +2,123 @@ package it.unibo.alchemist.test
...
+     */
+    fun <T, P : Position2D<P>> assertPedestriansReached(
+        env: Environment<T, P>,
+        tolerance: Double,
+        vararg coords: Number
+    ) {
+        val target = env.makePosition(*coords)
+        env.nodes
+            .filterIsInstance<OrientingPedestrian<T, *, *, *, *>>()
+            .forEach { p ->
```
DanySK, line 31: one liner

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/environments/ImageEnvironmentWithGraph.kt
```diff
@@ -9,12 +9,13 @@
 
 package it.unibo.alchemist.model.implementations.environments
 
-import it.unibo.alchemist.model.implementations.geometry.euclidean.twod.MutableConvexPolygonImpl
+import it.unibo.alchemist.model.implementations.geometry.euclidean.twod.AwtMutableConvexPolygon
```
DanySK, line 12: Why is the implementation referenced?

DanySK: ok found the reason

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/DoubleExtension.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package it.unibo.alchemist.model.implementations.geometry
+
+import org.danilopianini.lang.MathUtils
+import java.lang.IllegalArgumentException
+
+/**
+ * Maps a value in [min, max] to [newMin, newMax].
+ */
+fun Double.map(min: Double, max: Double, newMin: Double, newMax: Double): Double = when {
```
DanySK, line 18: What does it mean?

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/RangeExtension.kt
```diff
@@ -0,0 +1,98 @@
...
+    others.isEmpty() -> listOf(this)
+    else -> (this - others.first()).flatMap { it.subtractAll(others.drop(1)) }
+}
+
+/**
+ * Given a non empty list of points represented as vectors, this method finds the extreme
+ * coordinates (i.e. min and max coordinates) either on the X-axis or the Y-axis. [getXCoords]
+ * indicates which coordinates to extract, these are used to create the returned [ClosedRange].
+ */
+fun <V : Vector2D<V>> List<V>.findExtremeCoords(getXCoords: Boolean): ClosedRange<Double> {
```
DanySK, line 92: make this private, create a function for x and one for y: booleans do not map well to axes in an API

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/euclidean/twod/AbstractConvexPolygon.kt
```diff
@@ -0,0 +1,189 @@
...
+            else -> intersections.mapNotNull { it.point.orElse(null) }.distinct().size > 1
+        }
+    }
+
+    override fun closestEdgeTo(segment: Segment2D<Euclidean2DPosition>): Segment2D<Euclidean2DPosition> =
+        edges().minWith(
+            compareBy({
+                it.distanceTo(segment)
+            }, {
+                min(segment.distanceTo(it.first) + segment.distanceTo(it.second),
```
DanySK, line 65: make this three lines: first compute the arguments, then call min

alchemist/alchemist-implementationbase/src/main/java/it/unibo/alchemist/model/implementations/positions/Euclidean2DPosition.java
```diff
@@ -66,6 +66,24 @@ public Euclidean2DPosition normal() {
...
+    }
+
+    @NotNull
+    @Override
+    public Euclidean2DPosition coerceAtLeast(final double minimumMagnitude) {
+        if (getMagnitude() < minimumMagnitude) {
+            return resized(minimumMagnitude);
+        }
+        return this;
+    }
```
DanySK, line 85: That's cut/paste :)
and what does coercing mean for a position2D? Is it a method of Vector? Where is it defined?

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/DynamicPursuing.kt
paagamelo: `clear` would cause `volatileMemory[anyArea]` to be `null`, the operation performed sets `volatileMemory[anyArea]` to zero. It's useful to distinguish them because the pedestrian tries to avoid impasses for which `volatileMemory[area] != null` (= known impasses). Hence `clear` means the pedestrian would forget impasses as well, this is explained in the api doc of that method.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/ReachDestination.kt
paagamelo: In the class above, I'll move it to that class and reduce parametrisation.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/DynamicPursuing.kt
DanySK: Got it. Paste your own comment in the code, so that it's clarified forever!

paagamelo: Ok!

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/KnownDestinationReaching.kt
paagamelo: I moved both functions in the class above. 
I would avoid making `findKnownPath` an extension method of `P`, its signature now is very clean and I prefer having `findKnownPathTo(destination)` than `destination.knownPathTo()`.

DanySK: okay

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/NavigationPrioritisedSteering.kt
paagamelo: Factored it out, even though the function has generics and Kotlin doesn't support generic method reference.
Alternatively I may create a `PrevalentActionSelectionStrategy` functional interface and just go `PrevalentActionSelectionStrategy { ... }`, this may be clearer.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/utils/LazyMutable.kt
paagamelo: Yes

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/DoubleExtension.kt
paagamelo: Maps a value falling in range [min, max] to range [newMin, newMax], is this clearer?

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/RangeExtension.kt
paagamelo: Ok

alchemist/alchemist-implementationbase/src/main/java/it/unibo/alchemist/model/implementations/positions/Euclidean2DPosition.java
paagamelo: It's defined in `Vector`, I would have implemented it right there but there's a type problem with the curiously recurring template pattern: in `Vector<S : Vector<S>>` the return type of that method is `S`, but if I return `this` I get a `Vector<S>` and not a `S`, even though `S : Vector<S>`.

DanySK: Because this is a `Vector<S>`, not an `S`, for the compiler :)
anyway, since at some point this is forced to be `S`, I believe you can safely cast it.

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/DoubleExtension.kt
DanySK: No, I do not understand what it does. Can you provide an example? So maybe I can come up with some name

paagamelo: This is similar to [Arduino's map](https://www.arduino.cc/reference/en/language/functions/math/map/), it's a linear interpolation. Suppose you have an analog reading (= a value in [0, 1023] in Arduino boards) and you want to map it in the range [0, 180], representing the movement degree of a servomotor. So for instance 0 is mapped to 0, 1023 is mapped to 180 and 512 is mapped to 90.
An alternative name for the function is `linearlyInterpolate`.

alchemist/alchemist-implementationbase/src/main/java/it/unibo/alchemist/model/implementations/positions/Euclidean2DPosition.java
paagamelo: Ok, I'll have to suppress the unchecked cast warning as well.
By the way, we have two more pending conversations: one is on the first block of changes, the other is on this one. You'll have to click "Load more" to show them up.

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/DoubleExtension.kt
DanySK: Ah! Now I get it!
Please do not use `map` alone, it really makes people think of the `map` operation on traversables or of `let`.
It's not a linear interpolation either.

What you want to do is a *scaling* operation. Proposal: use `rescaleFromTo`. If you want it to be beautiful you may do something like:

```kotlin
internal data class DoubleInRange(val value: Double, val range: ClosedFloatingPointRange<Double>) {
   infix fun rescaledTo(newRange: ClosedFloatingPointRange<Double>) = TODO()
}
infix fun Double.within(range: ClosedFloatingPointRange<Double>) = DoubleInRange(this, range) // Plus check if it's in range and fail otherwise
// You should then be able to write something like
1 within 0..2 rescaledTo 2..4 // should return 3
```

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/NavigationPrioritisedSteering.kt
DanySK: Pick the one that you deem more readable

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/RangeExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+fun <T : Comparable<T>> rangeFromUnordered(bound1: T, bound2: T): ClosedRange<T> =
+    ObjectUtils.min(bound1, bound2)..ObjectUtils.max(bound1, bound2)
+
+/**
+ * Checks whether two ranges coincide. It's a different way of checking if they're equals,
+ * which doesn't depend on their actual implementation. Note that the way you obtain a range
+ * may influence the actual class used, and you can have two coincident ranges which don't
+ * result equals because of different classes.
+ */
+fun <T : Comparable<T>> ClosedRange<T>.coincide(other: ClosedRange<T>): Boolean =
```
DanySK, line 27: Uhmm... maybe `isCoincidentWith` reads better

DanySK: Or, `coincides`, at least. The way it's written now sounds like an active action

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/RangeExtension.kt
```diff
@@ -0,0 +1,110 @@
...
+ * Finds the intersection between two ranges which are guaranteed to intersect.
+ */
+private fun <T : Comparable<T>> ClosedRange<T>.unsafeIntersection(other: ClosedRange<T>): ClosedRange<T> =
+    ObjectUtils.max(start, other.start)..ObjectUtils.min(endInclusive, other.endInclusive)
+
+/**
+ * Finds the intersection between two ranges, the resulting range may feature a single value
+ * (if the ranges only share an endpoint) or can be null, if they don't intersect at all.
+ */
+fun <T : Comparable<T>> ClosedRange<T>.intersection(other: ClosedRange<T>): ClosedRange<T>? = when {
```
DanySK, line 53: `intersect`

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/euclidean/twod/AbstractConvexPolygon.kt
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    override fun closestEdgeTo(segment: Segment2D<Euclidean2DPosition>): Segment2D<Euclidean2DPosition> =
+        edges().minWith(
+            compareBy({
+                it.distanceTo(segment)
+            }, {
+                val segmentDistanceToEdgeBounds = segment.distanceTo(it.first) + segment.distanceTo(it.second)
+                val edgeDistanceToSegmentBounds = it.distanceTo(segment.first) + it.distanceTo(segment.second)
+                min(segmentDistanceToEdgeBounds, edgeDistanceToSegmentBounds)
```
DanySK, line 67: This lambda could become a private extension function of `Segment2D`, possibly in the companion

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/euclidean/twod/AbstractConvexPolygon.kt
```diff
@@ -0,0 +1,190 @@
...
+        return when {
+            intersections.any { it.type == SegmentsIntersectionType.SEGMENT } -> false
+            else -> intersections.mapNotNull { it.point.orElse(null) }.distinct().size > 1
+        }
+    }
+
+    override fun closestEdgeTo(segment: Segment2D<Euclidean2DPosition>): Segment2D<Euclidean2DPosition> =
+        edges().minWith(
+            compareBy({
+                it.distanceTo(segment)
```
DanySK, line 63: If the next lambda gets compacted, then this might become a one liner

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/DoubleExtension.kt
paagamelo: Very nice suggestion. However, I noticed that function is unused 😅 I'll remove it.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/KnownDestinationReaching.kt
```diff
@@ -0,0 +1,149 @@
...
+            cartesianProduct(landmarksAdjacentToStartRoom, landmarksAdjacentToEndRoom)
+    }
+
+    /**
+     * A [waypoint] is considered reached when inside [currentRoom] or in an adjacent room.
+     */
+    override fun isReached(waypoint: Euclidean2DPosition, currentRoom: ConvexPolygon): Boolean = with(action) {
+        super.isReached(waypoint, currentRoom) || doorsInSight().map { it.head }.any { it.contains(waypoint) }
+    }
+}
```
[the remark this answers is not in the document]
paagamelo: Do you refer to the documentation or the actual comments in the code?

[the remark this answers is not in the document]
DanySK: Forget it, I've been deceived by a GH highlighting bug

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/RouteFollowing.kt
```diff
@@ -0,0 +1,93 @@
...
+        actualNewRoom: ConvexPolygon
+    ) = with(action) {
+        doorsInSight()
+            .filter { it.head == previousRoom }
+            .minBy { it.distanceToPedestrian() }
+            ?.let { crossDoor(it) }
+            ?: inNewRoom(actualNewRoom)
+    }
+
+    override fun inNewRoom(newRoom: ConvexPolygon) = when {
```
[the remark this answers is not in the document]
paagamelo: I'll use nested ifs.

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/RangeExtension.kt
paagamelo: I agree, I'll combine the suggestions and use `coincidesWith`.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/NavigationPrioritisedSteering.kt
```diff
@@ -0,0 +1,62 @@
...
+        pedestrian,
+        { this.getSingleNavigationAction() },
+        maxWalk = { pedestrian.speed() / timeDistribution.rate },
+        toleranceAngle = Math.toRadians(toleranceAngle),
+        alpha = alpha
+    )
+) {
+
+    companion object {
+        private fun <T, M : ConvexPolygon> List<SteeringAction<T, Euclidean2DPosition>>.getSingleNavigationAction():
```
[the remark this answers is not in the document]
paagamelo: Yes, even though I'll need two type aliases to move the signature on a single line (maybe an overkill?)

alchemist/alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/geometry/euclidean/twod/AbstractConvexPolygon.kt
paagamelo: I refactored it in two private extension functions in `AbstractConvexPolygon`'s companion.

paagamelo:
@DanySK We have three open conversations (one on each "correction block"), you'll have to click "Load more" to show them up.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/NavigationPrioritisedSteering.kt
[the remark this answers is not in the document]
DanySK: do what you think is more readable

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/RouteFollowing.kt
```diff
@@ -0,0 +1,89 @@
...
+            super.inNewRoom(newRoom)
+        }
+    }
+
+    /**
+     * Checks if a new waypoint has been reached and updates [indexOfNextWaypoint] accordingly.
+     */
+    protected fun waypointReached(currentRoom: ConvexPolygon): Boolean = false
+        .takeIf { indexOfNextWaypoint >= route.size - 1 }
+        ?: route.indexOfLast { isReached(it, currentRoom) }.let { indexOfFartherWaypointReached ->
```
DanySK, line 76: let on a new line, plus I think this can get rewritten more cleanly using `also` to encapsulate the side effect

DanySK:
We should have two pending conversations now, right?
Plus: always keep the branch up to date with develop

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/NavigationPrioritisedSteering.kt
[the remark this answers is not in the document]
paagamelo: Ok I'll use type aliases.

alchemist/alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/navigationstrategies/RouteFollowing.kt
paagamelo: I rewrote it using more variables instead of `also`:
```kotlin
if (indexOfNextWaypoint >= route.size - 1) {
    return false
}
val indexOfFartherWaypointReached = route.indexOfLast { isReached(it, currentRoom) }
val newWaypointReached = indexOfFartherWaypointReached >= indexOfNextWaypoint
if (newWaypointReached) {
    indexOfNextWaypoint = // update it
}
return newWaypointReached

```
 I think this is the most readable way of writing it.

paagamelo:
@DanySK We have two pending conversations: the last one and another one on the first correction block (that's before codecov report), concerning the addition of `@JvmOverloads`annotation. I think the latter can be closed, you were concerned about the removal of the annotation but I added it (GH highlighting bug?).

DanySK:
Then we're done.

PR #422, opened 2020-06-11
Check all the generated fatJar

Placu95:
executes each generated fatJar with the option "--help" and checks its output (stdout and stderr) looking for not desired keywords.

Now the shadowJar tasks are executed after all check tasks are passed.
This avoids retrying the check tasks when the failure is in the shadowJar ones.


alchemist/buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -153,7 +153,7 @@ object Libs {
     /**
      * http://logback.qos.ch
      */
-    const val logback_classic: String = "ch.qos.logback:logback-classic:_"
+    const val logback_classic: String = "ch.qos.logback:logback-classic:1.2.3"
```
DanySK, line 156: This breaks versioning. Roll back and simply change the version in the `versions.properties` file

alchemist/buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -260,7 +260,7 @@ object Libs {
     /**
      * http://www.slf4j.org
      */
-    const val slf4j_api: String = "org.slf4j:slf4j-api:_"
+    const val slf4j_api: String = "org.slf4j:slf4j-api:1.7.30"
```
DanySK, line 263: Same as above

alchemist/buildSrc/src/main/kotlin/Libs.kt
Placu95: Done

alchemist/buildSrc/src/main/kotlin/Libs.kt
Placu95: Done

PR #545, opened 2020-09-23
Physical Agents Introduction and Cognitive Agents Refactoring

mazzio97:
Physical Agents are at an early stage but can be already used and tested for future works.
The cognitive model does not rely solely on IMPACT, which means is way more extensible now.
I have added a new module (alchemist-test) for testing purposes since the same methods were replicated across agents' modules.


DanySK:
@mazzio97 @paagamelo this currently has conficts, can you sync with develop?

mazzio97:
No problem, I will take care of it in a minute.

DanySK (approved):
Good stuff, congrats @paagamelo @mazzio97.
It's rare for me accepting as-is :+1:

PR #714, opened 2021-02-03
Update README

aPlacuzzi:
* Update `git clone` instruction with submodules
* Add fork instruction to specify that a secret with name `AUTOMERGE_TOKEN` is required by `automerge` action

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,13 @@ Contributions to this project are welcome. Just some rules:
 * Stay in sync with the `develop` branch: pull often from `develop` (if the build passes), so that you don't diverge too much from the main development line.
 * Do not introduce low quality code. All the new code must comply with the checker rules (that are quite strict) and must not introduce any other warning. Resolutions of existing warnings (if any is present) are very welcome instead.
 
+#### Forking the project
+This project defines several CI jobs. One of them (`automerge`) requires a secret with:
+
+* name = AUTOMERGE_TOKEN
+* value = a github token with `public_repo` scope
+
+If you don't provide it, then the CI workflow will fail.
 
 #### Building the project
 While developing, you can rely on IntelliJ to build the project, it will generally do a very good job.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -165,6 +165,13 @@ Contributions to this project are welcome. Just some rules:
 * Stay in sync with the `develop` branch: pull often from `develop` (if the build passes), so that you don't diverge too much from the main development line.
 * Do not introduce low quality code. All the new code must comply with the checker rules (that are quite strict) and must not introduce any other warning. Resolutions of existing warnings (if any is present) are very welcome instead.
 
+#### Forking the project
+This project defines several CI jobs. One of them (`automerge`) requires a secret with:
+
+* name = AUTOMERGE_TOKEN
+* value = a github token with `public_repo` scope
```
DanySK, line 172: Add a link for the token generation, should be portable.

README.md
```diff
@@ -165,6 +165,13 @@ Contributions to this project are welcome. Just some rules:
 * Stay in sync with the `develop` branch: pull often from `develop` (if the build passes), so that you don't diverge too much from the main development line.
 * Do not introduce low quality code. All the new code must comply with the checker rules (that are quite strict) and must not introduce any other warning. Resolutions of existing warnings (if any is present) are very welcome instead.
 
+#### Forking the project
```
DanySK, line 168: Move this before "Importing" and explain why one should fork instead of cloning directly

README.md
```diff
@@ -165,6 +165,13 @@ Contributions to this project are welcome. Just some rules:
...
+This project defines several CI jobs. One of them (`automerge`) requires a secret with:
+
+* name = AUTOMERGE_TOKEN
+* value = a github token with `public_repo` scope
+
+If you don't provide it, then the CI workflow will fail.
```
DanySK, line 174: ...even though the project status is correct.
So I'd say that the Workflow may appear as failed even though the build process is working correctly

DanySK (changes requested):

README.md
aPlacuzzi: Done.

README.md
aPlacuzzi: Done.

README.md
aPlacuzzi: Done.

PR #795, opened 2021-04-30
Fix some broken links in documentation

cric96:


DanySK:
Good to go as soon as we get CI confirmation that nothing's broken

PR #802, opened 2021-05-11
Scafi clone node

aPlacuzzi:
Support cloning of `ScafiNode` with the same behaviour of `ProtelisNode`


DanySK (changes requested):
A couple of minors

PR #856, opened 2021-07-08
Develop

kelvin-olaiya:
Some minor corrections in the  simulation configuration yaml on the wiki


DanySK:
Ideally, we should move all these examples into working test cases, and then load the files in the wiki from the test directly.
This way, the wiki would always be up to date.

Anyway, merging.

PR #858, opened 2021-07-12
Move cognitive pedestrian yaml example to website-snippets folder

kelvin-olaiya:
Before this PR is merged i must point out that not all
test cases i moved into the website-snippet folder passes. 
On the one hand running the normaly working configurations as they are now 
will produce this [error-log](https://github.com/AlchemistSimulator/Alchemist/files/6798910/test-error-log.txt), on the other hand the configuration regarding cognitive oriented pedestrian 
won't work even if runned normaly in the simulator because it seems that the node type is not inheriting from a `CognitivePedestrian` therefore incompatible with the `CognitiveBehavior` reaction. [log](https://pastebin.com/nFznWWVx)


DanySK:
Very good. Next step: we need to fix the tests.
The first issue that I see is that these classes cant be found on the classpath: we need to add `alchemist-cognitive-agents` to the `testRuntime` configuration of the test executor, in `build.gradle.kts`.
If you don't know how to do it I can provide further guidance. Let me know

kelvin-olaiya:
I've added the required configuration. Now the test produces the error output in the latter link
on the first comment

versions.properties
```diff
@@ -79,6 +79,8 @@ version.io.gitlab.arturbosch.detekt..detekt-formatting=1.17.1
 
 version.io.jenetics..jpx=2.2.0
 
+version.it.unibo.alchemist=11.3.0
```
DanySK, line 82: Why? This smells like a circular dependency

build.gradle.kts
```diff
@@ -107,6 +107,8 @@ allprojects {
         testImplementation("com.nhaarman.mockitokotlin2:mockito-kotlin:_")
         // Test runtime: Junit engine
         testRuntimeOnly(junit("engine"))
+        // Test runtime: Alchemist cognitive agents
+        testRuntimeOnly("it.unibo.alchemist:alchemist-cognitive-agents:_")
```
DanySK, line 111: Ah, I see. Please change as follows:
`testRuntimeOnly(alchemist("cognitive-agents"))`

otherwise, it will refer to a published artifact instead of accessing the subproject. Also, this must **not** be under `allprojects`, it's a dependency of the parent project.

DanySK (changes requested):
There is a mistake on dependencies

DanySK:
OK I see. `CognitivePedestrian` is not in the hierarchy of `CognitiveOrientingPedestrian2D`.
Direct delegation won't work. I fear we will need some boilerplate...

Try that in `CognitiveOrientingPedestrian2D`:

```kotlin
class CognitiveOrientingPedestrian2D<T, N : ConvexPolygon, E> @JvmOverloads constructor(
    environment: EuclideanPhysics2DEnvironmentWithGraph<*, T, N, E>,
    randomGenerator: RandomGenerator,
    knowledgeDegree: Double,
    group: PedestrianGroup2D<T>? = null,
    age: Age,
    gender: Gender,
    danger: Molecule? = null,
    private val consciousness: CognitivePedestrian2D<T> =
        CognitivePedestrian2D(environment, randomGenerator, age, gender, danger, group)
) : HomogeneousOrientingPedestrian2D<T, N, E>(
    environment,
    randomGenerator,
    knowledgeDegree = knowledgeDegree,
    group = group
),
    CognitivePedestrian<T, Euclidean2DPosition, Euclidean2DTransformation> {
```

and inside the class:

```kotlin
    override val cognitive: CognitiveModel get() = consciousness.cognitive
    override fun influencialPeople(): List<CognitiveAgent> = consciousness.influencialPeople()
    override val age: Age get() = consciousness.age
    override val gender: Gender get() = consciousness.gender
    override val compliance: Double get() = consciousness.compliance
    override fun probabilityOfHelping(
        toHelp: HeterogeneousPedestrian<T, Euclidean2DPosition, Euclidean2DTransformation>
    ): Double = consciousness.probabilityOfHelping(toHelp)
```

We are basically resorting to manual delegation, as some methods would conflict.

kelvin-olaiya:
Ok, now test are passing

DanySK (approved):

DanySK:
Everything looks good. Let us wait for the CI to ensure that everything is fine, then I'll rebase-merge.

PR #870, opened 2021-09-05
Implement pushing behavior on CognivitivePhysicalPedestrians

kelvin-olaiya:


alchemist-physical-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/nodes/CognitivePhysicalPedestrian2D.kt
```diff
@@ -28,13 +29,30 @@ open class CognitivePhysicalPedestrian2D<T> @JvmOverloads constructor(
     group: PedestrianGroup2D<T>? = null
 ) : CognitivePedestrian2D<T>(environment, randomGenerator, age, gender, danger, group), PhysicalPedestrian2D<T> {
 
-    /**
-     * TODO(maybe introduce some randomness? To model differences in people)
+    /*
+        According to [the work of Pelechano et al](https://bit.ly/3e3C7Tb) in order to bring out
+        pushing behavior different nodes must have different personal space threshold.
```
DanySK, line 34: Use the same multiline comment we use elsewhere:
```kotlin
/*
 * text goes here after the '*'
 */
```

DanySK (changes requested):
I am a bit puzzled by the fact that there are 42 commits, but very marginal changes and no additional tests. Is this a work in progress?
Since changes are pointwise, it would be fine to me to squash this in the mainline quickly, and then get on with the work with a simplified history.

kelvin-olaiya:
Many of those commits were from previous pull requests. The ones regarding this PR are the last 14.

Sorry but i don't quite understand what you mean by squashing in the mainline and simplifing the history.

DanySK:
Squashing means making all those commits a single commit. There are three techniques for unifying diverged branches:
* merging (creates a new commit, preserves the development history)
* squashing (compacts all commits of a pull request into a single commit)
* rebasing (preserves all commits, but re-arranges them in such a way that they build a linear history)

If I squash, I would recommend you to "start fresh" and continue development from the `develop` branch of the truth repository (this one).

DanySK (approved):

DanySK:
I think squashing is the way to go here. If you wanted to summarize this PR with a single commit message, what'd you use?

kelvin-olaiya:
Alright, I didn't realize that squashing is a git operation.

PR #996, opened 2021-12-22
fix(incarnation-scafi): return null in createT instead of throwing exception

cric96:
ScaFi incarnation currently throws an exception in `createT` method. 

However, in this way, we cannot easily use some action (for instance, [MoveToTarget](https://github.com/AlchemistSimulator/Alchemist/blob/4c167bc9eb3f3925936f62b5a620e5f15deee037/alchemist-implementationbase/src/main/java/it/unibo/alchemist/model/implementations/actions/MoveToTarget.java#L32) that uses [FollowTarget](https://github.com/AlchemistSimulator/Alchemist/blob/4c167bc9eb3f3925936f62b5a620e5f15deee037/alchemist-implementationbase/src/main/java/it/unibo/alchemist/model/implementations/movestrategies/target/FollowTarget.java#L74) and it assumes that the target molecule could be null).

Furthermore, [ProtelisNode](https://github.com/AlchemistSimulator/Alchemist/blob/4c167bc9eb3f3925936f62b5a620e5f15deee037/alchemist-incarnation-protelis/src/main/java/it/unibo/alchemist/model/implementations/nodes/ProtelisNode.java#L56) implements this method returning null, so I think that also ScaFiNode should return null accordingly.


DanySK (approved):
Excellent work.

Issue #1002, opened 2021-12-26
CSVExport exports data do not follow the column header


cric96:
# Bug overview
`CSVExport` accepts a List of `Export`. Each of them can extract some data from the current state of Alchemist simulation.
To do that, `Export` implementations need to define `columnNames` and `extractData`. In particular, `extractData` is a Map that should link each column name with a value.
The problem here is that there is a mismatch in the file exported from CSVExport from column names and their value.
# Example
Suppose you have implemented an `Export` with this column name:
```
val columnNames: List<String> = (1 to 5).map(i => s"temperature-$i").asJava // scala language
```
And then you have implemented the `extractData` method returning a Map for each temperature:
```
def extractData[T](....) = {
   val temperature1 = "temperature-1" -> 10
   val rest = (2 to 5).map("i => s"temperature-$i" -> 0)
   val allData = temperature1 :: rest
   Map(allData:_*).asJava
}
```
I excepted that the CSV that use this kind of export is structured as:
```
temperature-1 temperature-2 temperature-3 temperature-4 temperature-5
10 0 0 0 0
```
But, the file produced is something like:
```
temperature-1 temperature-2 temperature-3 temperature-4 temperature-5
0 0 0 0 10
```
# How to reproduce

[Here](https://github.com/cric96/alchemist-bugs/tree/export-problem) there is an MWE where the problem is shown.
So:
1- clone the repository
2- run `./gradlew runAllBatch`
3- look at data/ to verify that the export file is effectively wrong

# Possible problem

`CSVExport` prints the first line using `columnNames` but then it takes the `extractData` result directly without following the same 
order of the `columnNames` value:

**column names part**
```
dataExtractors.flatMap {
    it.columnNames
}.forEach {
    outputPrintStream.print(it)
    outputPrintStream.print(" ")
}
```
**extract data part**
```
dataExtractors.forEach {
    it.extractData(environment, reaction, time, step).values.forEach { value ->
        outputPrintStream.print(value)
        outputPrintStream.print(' ')
    }
}
```

# Possible solution
Mapping the column names with the output produced by the extractor.
```
val data = extractor.extractData(environment, reaction, time, step)
extractor.columnNames.map { data[it] }.filterNotNull().forEach {
    value -> outputPrintStream.print(value)
    outputPrintStream.print(' ')
}
```
Probably it is not the best solution in Kotlin so far, but it is just to underline the problem.

DanySK:
[...]

PR #1005, opened 2021-12-27
test(loading): add regression test for #1002

cric96:


alchemist-loading/src/test/kotlin/it/unibo/alchemist/test/TestCSVExporter.kt
```diff
@@ -24,24 +26,53 @@ import java.io.File
...
+            ?.dropWhile { !it.contains("d c b a") } // remove the lines before the column names
+            ?.drop(1) // I am not interested in column head
+            ?.first()
+        exporterFirstLine.shouldNotBeNull()
+        exporterFirstLine.shouldNotBeEmpty()
+        exporterFirstLine.shouldContain("0 1 2 3")
+    }
+}) {
+    /* common utility functions */
+    companion object {
```
DanySK, line 56: These could actually be added to alchemist-test, or maybe they are already there in some form, and only need to get enhanced

PR #1006, opened 2021-12-27
test(loading): add reproduction test for #1002

cric96:


DanySK:
Uhmm... this is expected to fail, but it works already :open_mouth:

cric96:
Ops, I think that giving the same name at two different PRs was not a good idea :smile: 

Indeed, looking at the actions results, you will see that this PR is failed:
https://github.com/AlchemistSimulator/Alchemist/actions/runs/1628357671
and the other is successful:
https://github.com/AlchemistSimulator/Alchemist/actions/runs/1628357686

DanySK:
I think that we hit a bug in GitHub, the check suite tells me that tests are passing, but they are no. Anyway...

DanySK:
The parallel PR has been merged into #1004

PR #1318, opened 2022-06-01
feat(api)!: add global reactions

kelvin-olaiya:


alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java
```diff
@@ -345,12 +346,12 @@ public void play() {
     }
 
     @Override
-    public void reactionAdded(final Reaction<T> reactionToAdd) {
+    public void reactionAdded(final GlobalReaction<T> reactionToAdd) {
         reactionChanged(new ReactionAddition(reactionToAdd));
```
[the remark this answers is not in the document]
DanySK: We do need to fix this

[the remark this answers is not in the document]
DanySK: I reviewed this. `run` is protected by a lock on `environment`, and `myThread` is set once. calls to `reactionAdded` (hence, to `checkCaller`) ran before the assignment would compare with `null` and throw `IllegalMonitorStateException`. If the calling thread is not the simulation thread, we DO want the exception, so even if there is a race condition, it is fine (it would enter the condition anyway). For a race that produces unwanted results to occur (namely: the calling thread is the simulation thread, but an exception is raised), we would need the simulation thread to be both setting myThread and reading it, which is impossible.

I believe this is a false positive, though -- @kelvin-olaiya ? @cric96 @metaphori guys what do you think? Can you give us an opinion?

alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java
```diff
@@ -345,12 +346,12 @@ public void play() {
...
     }
 
     @Override
-    public void reactionRemoved(final Reaction<T> reactionToRemove) {
+    public void reactionRemoved(final GlobalReaction<T> reactionToRemove) {
         reactionChanged(new ReactionRemoval(reactionToRemove));
```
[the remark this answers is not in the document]
DanySK: Same as the previous, closing and waiting for comments.

alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java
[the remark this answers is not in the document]
cric96: TL;DR I check the code and I agree with @danysk: I think that this is a false positive.

The only potential harmful situation that popup in my mind is when you have a concurrent thread that tries to add remove/add a reaction before "myThread" is assigned (probably simulation thread is a better choice for the name??). Anyway, in this case, as @danysk says, the variable is still `null`, so the check caller will be thrown the exception, that is what I excepted will happen...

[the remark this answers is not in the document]
kelvin-olaiya: I also agree with both @DanySK and @cric96. The `checkCaller` call prevents the race condition. If any changes in the simulation must be performed by `myThread`,  any concurrent thread trying to remove/add reaction should and will throw the Exception. 
At this point, I also agree with @cric96 on the thread naming. I will rename it to `simulationThread` as soon I'll get back to work.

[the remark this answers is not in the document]
DanySK: Very good. Marking as solved.

DanySK:
Uhmm... I'm not convinced that:

```mermaid
classDiagram
GlobalReaction <|-- Reaction
```

A `Reaction` is **not** a special `GlobalReaction`
We probably need something like

```mermaid
classDiagram
WeNeedToNameThis <|-- Reaction
WeNeedToNameThis <|-- GlobalReaction
class WeNeedToNameThis {
          +List<Dependency> getInboundDependencies()
          +List<Dependency> getOutboundDependencies()
      }
class Reaction {
          +Node getNode()
          +Context getInputContext()
          +Context getOutputContext()
      }
```

This should be a sealed hierarchy:

```kotlin
sealed interface WeNeedToNameThis
```

So that in the dependency graph we can depend on the sub-interfaces directly.

kelvin-olaiya:
Yeah, I had this doubt at some point in the refactoring. Since I was trying to maintain `Reaction` as the interface name, I wrongly preferred technicality to design.
My suggestion for root interface could be:
* `AlchemistReaction`
* `GenericReaction`
* `BaseReaction`

DanySK:
And `getNode()`, I suppose.

It is a time-distributed object with dependencies and an execution strategy. I would use `Actionable`

kelvin-olaiya:
No, `getNode()` and related will remain in `Reaction`, otherwise `GlobalReaction` inheritors will have to have an empty or exception throwing method. Anyway, `Actionable` sounds reasonable.

DanySK:
I was continuing on your "execept": "except for those getting the in/out context and `getNode`" (as Node is not a property of `Actionable`.
No empty method:
```kotlin
sealed interface Actionable: Serializable
interface GlobalReaction : Actionable
interface Reaction: Actionable
```

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/PhysicsUpdate.kt
```diff
@@ -0,0 +1,59 @@
...
+     */
+    val environment: Dynamics2DEnvironment<T>,
+    override val timeDistribution: TimeDistribution<T>,
+) : GlobalReaction<T> {
+
+    override val outboundDependencies: ListSet<out Dependency> = ListSet.of()
+        get() = ImmutableListSet.copyOf(field)
+
+    override val inboundDependencies: ListSet<out Dependency> = ListSet.of(Dependency.MOVEMENT)
+        get() = ImmutableListSet.copyOf(field)
```
kelvin-olaiya, line 38: @DanySK, I'm a bit stuck sorting out how to properly manage the Dependency Graph.
In `PhysicalBlendedSteering`, I've added an outbound dependency on `MOVEMENT`, 
I'm I missing something? Because what I'm getting is that the local reactions are executed just once.

DanySK: If you can only move as a result of the action, then it is correct. If your reaction (or its conditions) *depends on* the movement of other nodes (namely, the reaction reads the position of others, and may be scheduled sooner or later depending on their position), then they must also declare an inbound reaction.

kelvin-olaiya: So, since the `GlobalReaction` reads the velocities of all the nodes in the environment, which they have been set by the local reactions, it should be:
* **Global reaction** with both an inbound and outbound MOVEMENT dependency
* **Local reaction** with just the outbound MOVEMENT dependency
Right?

kelvin-olaiya: No, still the Local reaction are executed once at the start of the simulation and never again. 
Also, the time remains still at `0.0` while the step counter increases.

DanySK: GlobalReaction is an interface. It has no context, it is always implicitly treated as GLOBAL, as LOCAL and NEIGHBORHOOD make no sense.

DanySK: It's scheduled as ASAP (rate to infinity) and dependencies say that it is executable, so it gets rescheduled continually.

kelvin-olaiya: > GlobalReaction is an interface. It has no context, it is always implicitly treated as GLOBAL, as LOCAL and NEIGHBORHOOD make no sense.

Yes sorry, with Global reaction i meant its implementation which is `PhysicsUpdate` and with local reaction i meant `PhysicalBlendedSteering`. 

I have already removed the context properties from the `GlobalReaction` interface. 

It's getting a little tricky for me... From what I'm getting now, having said that `PhysicsUpdate` has both in and out MOVEMENT dependency, in order to get all the `PhysicalBlendedSteering` to be executed, they should also have they inbound dependency. 

Did i miss something for the `PhysicsUpdate` having INFINITY rate? I did give it a time distribution.

DanySK: The rate is assigned by the incarnation, and Protelis defaults to "ASAP", I think. You should use a `DiracComb`. Also, I'm not sure the correct dependency is `MOVEMENT`. Can we discuss this briefly in person?

alchemist-api/src/main/kotlin/it/unibo/alchemist/model/interfaces/Actionable.kt
```diff
@@ -0,0 +1,103 @@
...
+
+package it.unibo.alchemist.model.interfaces
+
+import org.danilopianini.util.ListSet
+import java.io.Serializable
+
+/**
+ * A time-distributed entity with [inboundDependencies], [outboundDependencies] and an execution strategy.
+ */
+sealed interface Actionable<T> : Serializable {
```
DanySK, line 18: Actionable must be Comparable. Subclasses are comparable with Actionable.

alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java
```diff
@@ -186,49 +187,50 @@ private void doOnStatus(final Runnable fun) {
     }
 
     private void doStep() {
-        final Reaction<T> mu = scheduler.getNext();
-        if (mu == null) {
+        final Actionable<T> reaction = scheduler.getNext();
```
DanySK, line 190: `reaction` $\to$ `nextEvent`

alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java
```diff
@@ -186,49 +187,50 @@ private void doOnStatus(final Runnable fun) {
...
         } else {
-            final Time t = mu.getTau();
-            if (t.compareTo(currentTime) < 0) {
-                throw new IllegalStateException(mu + "\nis scheduled in the past at time " + t
+            final Time scheduledTime = nextEvent.getTau();
+            if (scheduledTime.compareTo(currentTime) < 0) {
+                throw new IllegalStateException(nextEvent + "\nis scheduled in the past at time " + scheduledTime
                         + ", current time is " + currentTime
                         + ". Problem occurred at step " + currentStep);
             }
```
DanySK, line 200:
```suggestion
                throw new IllegalStateException(
                    nextEvent + " is scheduled in the past at time " + scheduledTime
                        + ", current time is " + currentTime
                        + ". Problem occurred at step " + currentStep
                );
            }
```

alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java
```diff
@@ -590,29 +595,34 @@ private NodeAddition(final Node<T> source) {
...
-        private final Reaction<T> actualSource;
+        private final Actionable<T> actualSource;
 
         private UpdateOnReaction(final Reaction<T> source) {
             super(source.getNode());
             actualSource = source;
         }
 
+        private UpdateOnReaction(final Actionable<T> source) {
+            super(null);
```
DanySK, line 606: This is a call for NPEs, plus constructor calls are linked at compile-time... We need to revise the `Update` interface

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/PhysicsDependency.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+enum class PhysicsDependency : Dependency {
+
+    /**
+     * A dependency on any physical update in the world.
+     */
+    PHYSICS;
+
+    override fun toString(): String = "Physics Update"
+}
```
DanySK, line 25:
```suggestion
/**
 * A dependency on updates of the physics engine.
 */
object PhysicsDependency : Dependency {
    override fun toString(): String = "Physics Update"
}
```

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/PhysicalBlendedSteering.kt
```diff
@@ -40,6 +41,10 @@ class PhysicalBlendedSteering<T>(
     override val steerStrategy: SteeringStrategy<T, Euclidean2DPosition> =
         Sum(environment, node, super.steerStrategy)
 
+    init {
+        addOutboundDependency(PhysicsDependency.PHYSICS)
```
DanySK, line 45:
```suggestion
        addOutboundDependency(Physics)
```

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/PhysicsUpdate.kt
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * A global Reaction responsible for updating the physics of an [Dynamics2DEnvironment].
+ */
+class PhysicsUpdate<T> @JvmOverloads constructor(
+    /**
+     * The environment to update.
+     */
+    val environment: Dynamics2DEnvironment<T>,
+    updateRate: Double = 10.0,
```
DanySK, line 33: why this default?

alchemist-cognitive-agents/src/main/kotlin/it/unibo/alchemist/model/implementations/actions/AbstractSteeringAction.kt
```diff
@@ -39,10 +39,12 @@ abstract class AbstractSteeringAction<T, P, A>(
           P : Vector<P>,
           A : GeometricTransformation<P> {
 
+    private val pedestrian = node.asProperty<T, PedestrianProperty<T>>()
```
DanySK, line 42: why not just requiring pedestrian in the constructor, then?

alchemist-euclidean-geometry/src/main/kotlin/it/unibo/alchemist/model/implementations/environments/Continuous2DEnvironment.kt
```diff
@@ -176,19 +177,19 @@ open class Continuous2DEnvironment<T>(incarnation: Incarnation<T, Euclidean2DPos
      * Checks if a node doesn't overlap with any other node in the environment (see [overlappingNodes]). If the
      * node is shapeless, true is returned.
      */
-    private fun Node<T>.canFit(position: Euclidean2DPosition): Boolean =
-        asPropertyOrNull<T, AreaProperty<T>>() == null || overlappingNodes(position).isEmpty()
+//    private fun Node<T>.canFit(position: Euclidean2DPosition): Boolean =
+//        getShape(this) != adimensional || overlappingNodes(position).isEmpty()
```
DanySK, line 181: Remove these commented lines

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/properties/PhysicalPedestrian2D.kt
```diff
@@ -48,15 +49,18 @@ class PhysicalPedestrian2D<T>(
...
     override val comfortRay: Double get() {
-        val cognitiveModel = node.asPropertyOrNull<T, CognitiveProperty<T>>()?.cognitiveModel
         return if (cognitiveModel?.wantsToEscape() == true) {
             desiredSpaceTreshold / 3
         } else {
             desiredSpaceTreshold
         }
     }
 
+    private val fallenAgentListerners: MutableList<(Node<T>) -> Unit> = mutableListOf()
```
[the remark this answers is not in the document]
kelvin-olaiya: You mean that user should replace the list whenever they want to add a listener?

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/reactions/PhysicsUpdate.kt
kelvin-olaiya: I'll default to 1.0

alchemist-physics/src/main/kotlin/it/unibo/alchemist/model/implementations/properties/PhysicalPedestrian2D.kt
```diff
@@ -76,7 +80,10 @@ class PhysicalPedestrian2D<T>(
     private val Node<T>.position get() = environment.getPosition(this)
 
     override fun checkAndPossiblyFall() {
-        isFallen = isFallen || shouldFall(repulsionForces())
+        if (!isFallen && shouldFall(repulsionForces())) {
+            isFallen = true
+            fallenAgentListeners.forEach { it.invoke(node) }
```
DanySK, line 85:
```suggestion
            fallenAgentListeners.forEach { hasFallen -> hasFallen(node) }
```

DanySK (changes requested):

DanySK:
@kelvin-olaiya let us first cleanup the log. There are missing `toString` that make the error hard to read:

```


* GenericNode -> Node0{ properties: [PtDevice0, it.unibo.alchemist.model.implementations.properties.Human@56f1ed94, it.unibo.alchemist.model.implementations.properties.Perceptive2D@79844d7e, it.unibo.alchemist.model.implementations.properties.CognitivePedestrian@440ca570, it.unibo.alchemist.model.implementations.properties.Cognitive2D@79188289, it.unibo.alchemist.model.implementations.properties.CircularArea@205c5d73], molecules: {} } |  
-- | --
  | * Node -> Node0{ properties: [PtDevice0, it.unibo.alchemist.model.implementations.properties.Human@56f1ed94, it.unibo.alchemist.model.implementations.properties.Perceptive2D@79844d7e, it.unibo.alchemist.model.implementations.properties.CognitivePedestrian@440ca570, it.unibo.alchemist.model.implementations.properties.Cognitive2D@79188289, it.unibo.alchemist.model.implementations.properties.CircularArea@205c5d73], molecules: {} } |  
  | * ProtelisDevice -> PtDevice0 |  
  | * Human -> it.unibo.alchemist.model.implementations.properties.Human@56f1ed94 |  
  | * Perceptive2D -> it.unibo.alchemist.model.implementations.properties.Perceptive2D@79844d7e |  
  | * CognitivePedestrian -> it.unibo.alchemist.model.implementations.properties.CognitivePedestrian@440ca570 |  
  | * Cognitive2D -> it.unibo.alchemist.model.implementations.properties.Cognitive2D@79188289 |  
  | * CircularArea -> it.unibo.alchemist.model.implementations.properties.CircularArea@205c5d73
```

So please add a reasonable toString to all of these properties.

Then, the source of the error is pretty clear: there is no registered `PedestrianProperty`: does `CognitivePedestrian` implement it?

DanySK:
Found it: https://github.com/kelvin-olaiya/Alchemist/blob/e7629be9f06189428d060ce7aa6c07bff8cd4463/alchemist-loading/src/main/kotlin/it/unibo/alchemist/loader/m2m/LoadingSystem.kt#L228 should be done through https://github.com/kelvin-olaiya/Alchemist/blob/e7629be9f06189428d060ce7aa6c07bff8cd4463/alchemist-loading/src/main/kotlin/it/unibo/alchemist/loader/m2m/LoadingSystem.kt#L280 to register the whole hierarchy.

It may make sense to rename that method to `contextualize`, then create a second method (wrapping factory.deregistersingleton) called `decontextualize`. That would make the loading clearer, providing a facade on `factory`.

DanySK:
Of course it is :+1:

DanySK (approved):
Looks good

PR #1377, opened 2022-06-27
chore: add toString() to all NodeProperties

kelvin-olaiya:


DanySK:
Node ids are not particularly interesting (they are properties of the node). I suggest to show the (Kotlin) properties of the actual NodeProperties. If there's none (tag properties) they can be just the name of the class.

DanySK:
Further reccomendation: create an `AbstractNodeProperty` that all properties inherti from with this toString:

```kotlin
override fun toString(): String = with(this::class) { simpleName ?: qualifiedName ?: jvmName }
```

DanySK:
Commit [20f50a3](https://github.com/AlchemistSimulator/Alchemist/pull/1377/commits/20f50a3532951c4d045eded2dbff3bcdb7b196f6) is very unrelated, can you open another PR for that one and rebase it out of this one?

DanySK:
Good to go. Enabled autosquash.

PR #1379, opened 2022-06-27
fix(alchemist-loading): fix properties singleton registration

kelvin-olaiya:


DanySK (changes requested):
there must be **no** call to `factory.deRegisterSingleton` but in `decontextualize`.

PR #1381, opened 2022-06-28
test(incarnation-protelis): check that env.get("layer name") never returns null, fails if no layer is available.

cric96:


DanySK:
This test [fails correctly](https://scans.gradle.com/s/ofr6yztmgmef2). I'll fix this shortly and then merge the fix in. I need to fix the releases first.

PR #1384, opened 2022-06-30
fix(implementationbase): fix errors on programmatic node removal via the `RemoveNode` `Action`

cric96:
I try to add action to remove nodes from the environment (i.e., `RemoveNode` right?):
```
deployments:
  type: Grid
  parameters: [-5, -5, 5, 5, 0.5, 0.5, 0.25, 0.25]
  programs:
    - time-distribution: 1
      type: Event
      actions:
      - type: RemoveNode
```
But it throws the following exception (more details in the GHA [action log](https://github.com/AlchemistSimulator/Alchemist/runs/7128608634?check_suite_focus=true#step:3:1577)):
```
java.lang.IllegalArgumentException: no such vertex in graph: Event@2.0011337332612453:[]-1.0->[RemoveNode]
```
Probably related to the fact that the engine tries to access the outbound dependencies of a reaction that does not exist anymore:
https://github.com/AlchemistSimulator/Alchemist/blob/580fbf94d20b39e016c1653d734926dd41642125/alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java#L206-L213


DanySK:
Oh, god, no... This is a **major** regression

alchemist-full/src/test/resources/remove.yml
```diff
@@ -0,0 +1,13 @@
...
+
+_remover: &remover
+  - time-distribution: 1
+    type: Event
+    actions:
+      - type: RemoveNode
+
+deployments:
+  type: Grid
+  parameters: [-5, -5, 5, 5, 0.5, 0.5, 0.25, 0.25]
```
DanySK, line 11:
```suggestion
  parameters: [-1, -1, 1, 1, 0.5, 0.5, 0.25, 0.25]
```

Do we need such a big network? Using something smaller would speed up the tests.

DanySK:
@cric96 should you have a solution, please apply it. I got no time to work on this in the forthcoming days

cric96:
> @cric96 should you have a solution, please apply it. I got no time to work on this in the forthcoming days

@DanySK A way to solve this problem consists in checking if the reaction exists in the graph before computing the outbound dependencies.

I implement this solution [here](https://github.com/AlchemistSimulator/Alchemist/pull/1384/files#diff-e8aab2a2e92b7f79c03280a914cbd032ad37011178d165c61c63a23326e4bb64R185-R191)
Two points:
- is this the idiomatic way to implement using kotlin?
- the `outgoingEdgesOf` calls `assertVertexExist` that internally calls `containsVertex`. So, `outboundDependencies` call two times the `containsVertex` method. It could be a performance drawback? Should it is better to use a "try and catch" solution?

DanySK:
I'm not sure it's that simple. The action removes a node, but other actions may have been executed before. The dependencies of these previous actions need to be updated. The fix makes sense (I personally favor `when` over `if`, but whatever), but we need to ensure that we deal with dependency updates correctly.

DanySK:
@cric96 does removing nodes from the GUI work? select and then (I think) <kbd>D</kbd>

DanySK:
(without this patch, I mean)

cric96:
> @cric96 does removing nodes from the GUI work? select and then (I think) D

Yes, it works!

cric96:
> I'm not sure it's that simple. The action removes a node, but other actions may have been executed before. The dependencies of these previous actions need to be updated. The fix makes sense (I personally favor `when` over `if`, but whatever), but we need to ensure that we deal with dependency updates correctly.

Probably I am wrong, but you already take the dependency before of effectively removing the Reaction `mu` here: https://github.com/AlchemistSimulator/Alchemist/blob/580fbf94d20b39e016c1653d734926dd41642125/alchemist-engine/src/main/java/it/unibo/alchemist/core/implementations/Engine.java#L208
The problem is that, after performing `performChanges`, you definitively remove the Reaction in the graph, and therefore when you call `outboundDependecies` again the reaction does not exist anymore.

DanySK:
It does make sense, indeed

PR #1398, opened 2022-07-03
feat(loading): loading support for global reactions

kelvin-olaiya:


alchemist-api/src/main/java/it/unibo/alchemist/model/interfaces/Incarnation.java
```diff
@@ -75,16 +75,13 @@
...
      *            the environment that will host this object
-     * @param node
-     *            the node that will host this object
      * @param parameter
      *            a {@link String} describing the object
      * @return a new {@link TimeDistribution}
      */
     TimeDistribution<T> createTimeDistribution(
             RandomGenerator randomGenerator,
             Environment<T, P> environment,
```
DanySK, line 84: I think it would be better to make the parameter optional (aka: `@Nullable`) rather than removing it. The documented behaviour is: if it's null, it is a reaction not belonging to a `Node`). Otherwise, it is a "normal" reaction.

alchemist-incarnation-biochemistry/src/test/java/it/unibo/alchemist/test/TestBiomolLayer.java
```diff
@@ -53,7 +53,7 @@ void testBiomolStepLayer() {
         final Molecule a = new Biomolecule("A");
         final Reaction<Double> underTest = INCARNATION.createReaction(
             rand, environment, cellNode,
-            INCARNATION.createTimeDistribution(rand, environment, cellNode, "1"),
+            INCARNATION.createTimeDistribution(rand, environment, "1"),
```
DanySK, line 56: revert, as per previous discussion

DanySK (changes requested):

alchemist-api/src/main/java/it/unibo/alchemist/model/interfaces/Incarnation.java
kelvin-olaiya: But the method is for the creation of a `TimeDistribution` not a `Reaction`. The node parameter is never used.

DanySK: But it might. In general, a `TimeDistribution` might use information from the `Node` (in which case it'd be unusable for `GlobalReactions`). So the parameter must be there, annotated with `@Nullable`, and should not disappear (even if all the existing implementations do not use it).

kelvin-olaiya: Alright

alchemist-api/src/main/java/it/unibo/alchemist/model/interfaces/Incarnation.java
```diff
@@ -125,12 +126,37 @@ Reaction<T> createReaction(
...
-            TimeDistribution<T> time,
-            Reaction<T> reaction,
-            String additionalParameters
+        RandomGenerator randomGenerator,
+        Environment<T, P> environment,
+        Node<T> node,
+        TimeDistribution<T> time,
+        Reaction<T> reaction,
+        String additionalParameters
+    );
```
DanySK, line 135: This could be removed, I think

alchemist-api/src/main/java/it/unibo/alchemist/model/interfaces/Incarnation.java
```diff
@@ -149,12 +175,36 @@ Condition<T> createCondition(
...
-            Node<T> node,
-            TimeDistribution<T> time,
-            Reaction<T> reaction,
-            String additionalParameters
+        RandomGenerator randomGenerator,
+        Environment<T, P> environment,
+        Node<T> node,
+        TimeDistribution<T> time,
+        Reaction<T> reaction,
+        String additionalParameters
```
DanySK, line 183: This could be removed, I think

DanySK (changes requested):

DanySK (approved):

PR #1400, opened 2022-07-04
fix(physics): fix physics update rate

kelvin-olaiya:
I've found the reason why the simulation seemed to be slower after the global reaction introduction.
Here's the [link](https://github.com/dyn4j/dyn4j/issues/37) to the issue on the physics engine's repo with the explanation.


DanySK:
@kelvin-olaiya due to strict time constraints of your thesis work I merged this in immediately, but after your manuscript is submitted, please come back and try to devise a test :test_tube:

kelvin-olaiya:
Ok, sure!

PR #1402, opened 2022-07-05
fix(physics): make physics engine aware of obstacles

kelvin-olaiya:
In `EnvironmentWithDynamics` node position are computed by the physics engine. 
Thus, it has to know if there are any obstacles in the environment to detect collision with them.


DanySK:
@kelvin-olaiya we do need a test for this

PR #1452, opened 2022-07-21
feat(incarnation-scafi): add support for multiple Scafi programs at once (align `RunProgramExport` in the `Send` action using the program name)

cric96:


DanySK:
good stuff

PR #1646, opened 2022-11-02
fix: adjust padding in menu footer

cric96:
image


DanySK:
Lovely :love_you_gesture:

PR #1729, opened 2022-12-02
ci: add uploadJvm and uploadJs to release

AngeloFilaseta:
I did some extra researches just to be 100% sure and differently from what I specified in the email, uploadAll seems to include some fatJars. This is caused by uploadJavaOSSRH. Since we don't need this last task, I added the uploadJvm and uploadJs tasks instead.

### DIFF
--- a/alchemist-web-renderer/build.gradle.kts
+++ b/alchemist-web-renderer/build.gradle.kts
@@ -128,3 +128,13 @@ publishing.publications {
         }
     }
 }
+
+publishing {
+    publications {
+        publications.withType<MavenPublication>().configureEach {
+            if ("OSSRH" !in name) {
+                artifact(tasks.javadocJar)
+            }
+        }
+    }
+}
--- a/alchemist-web-renderer/build.gradle.kts
+++ b/alchemist-web-renderer/build.gradle.kts
@@ -128,13 +128,3 @@ publishing.publications {
         }
     }
 }
-
-publishing {
-    publications {
-        publications.withType<MavenPublication>().configureEach {
-            if ("OSSRH" !in name) {
-                artifact(tasks.javadocJar)
-            }
-        }
-    }
-}
--- a/release.config.js
+++ b/release.config.js
@@ -15,7 +15,8 @@ git -C build/website/ add . || exit 1
 git -C build/website/ commit -m "chore: update website to version \${nextRelease.version}" || exit 2
 git -C build/website/ push || exit 3
 ./gradlew shadowJar --parallel || ./gradlew shadowJar --parallel || exit 4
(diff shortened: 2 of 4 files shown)


DanySK:
This sounds good.

release.config.js
```diff
@@ -15,7 +15,7 @@ git -C build/website/ add . || exit 1
 git -C build/website/ commit -m "chore: update website to version \${nextRelease.version}" || exit 2
 git -C build/website/ push || exit 3
 ./gradlew shadowJar --parallel || ./gradlew shadowJar --parallel || exit 4
-./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform release --parallel || ./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform release --parallel || ./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform release --parallel exit 5
+./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform uploadJvm uploadJs release --parallel || ./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform uploadJvm uploadJs release --parallel || ./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform uploadJvm uploadJs release --parallel exit 5
```
DanySK, line 18: OK this is getting waaaay to long
```suggestion
RELEASE_ON_CENTRAL="./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform uploadJvm uploadJs release --parallel"
eval "$RELEASE_ON_CENTRAL" || eval "$RELEASE_ON_CENTRAL" || eval "$RELEASE_ON_CENTRAL" || exit 5
```

this is getting waaay too long otherwise... I hope it works als like this

alchemist-web-renderer/build.gradle.kts
```diff
@@ -128,3 +128,13 @@ publishing.publications {
...
+
+publishing {
+    publications {
+        publications.withType<MavenPublication>().configureEach {
+            if ("OSSRH" !in name) {
+                artifact(tasks.javadocJar)
+            }
+        }
+    }
+}
```
DanySK, line 140: This is okay, but maybe it's better in the main `build.gradle.kts`, applied to all "multiplatform" plugins? So that if we add another MP project its publication gets configured transparently. What do you think?

DanySK (changes requested):
We may consider adding two tasks:

`dryDeliveryToMavenCentral` which depends on `uploadKotlinOSSRH` `uploadKotlinMultiplatform` `uploadJvm` `uploadJs` `close`

and 

`deliveryToMavenCentral` which depends on `dryDeliveryToMavenCentral` and `release`

alchemist-web-renderer/build.gradle.kts
AngeloFilaseta: Oh, right, that's my bad

release.config.js
AngeloFilaseta: Is it ok if we try with this solution first? I think I like this solution a little bit more if I compare it to the task approach, as the root build.gradle.kts is already not-that-small and these commands are not used elsewhere.

DanySK: Yeah

DanySK (approved):

DanySK:
Commit history is clean enough, enabling auto-rebase-merge.

PR #1751, opened 2022-12-05
ci(web-renderer): test JS release on Maven Central

DanySK:


DanySK:
@AngeloFilaseta, this PR is meant to help debug why uploading the JS stuff made the release on Central fail.

AngeloFilaseta:
We can get rid of the `implementation(project.dependencies.enforcedPlatform(libs.kotlin.wrappers.bom))`  dependency from `alchemist-web-renderer` jsMain sourceSet. With just few changes everything works in the project and the CD should make it. The dependency can also be removed from the catalog and the `yarn.lock` file needs to be updated. Is it possible for me to update the current PR or do I need to create a new one?

DanySK:
You cannot update this PR unless I give you write access to the project/organization, which I prefer not to.

BUT you can open a PR having branch `ci/(web-renderer)-test-JS-release-on-Maven-Central` as base. Updates to that branch would update this PR, the execution would check that things work, and we should be good to go.

PR #1752, opened 2022-12-05
fix(deps): remove problematic dependency for Js release

AngeloFilaseta:


DanySK:
Merging immediately into the feature branch to trigger a branch build.

PR #1849, opened 2023-01-11
feat(web-renderer): add EnvironmentMonitor and ServerState

AngeloFilaseta:
This PR contains the EnvironmentMonitor implementation. The monitor maps the Environment to its surrogate class and stores it in the Server state, which is also added in this PR.

### DIFF
--- a/alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/state/ServerState.kt
+++ b/alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/state/ServerState.kt
@@ -18,8 +18,14 @@ import it.unibo.alchemist.server.state.reducers.simulationReducer
 
 /**
  * The state of the server.
+ * The [ServerState] is managed using the
+ * <a href="https://reduxkotlin.org/introduction/core-concepts">Core concepts of the ReduxKotlin library</a>.
+ * Like in the original Redux library the state is stored in a single class that contains other objects via composition.
+ * The state can be changed using actions that must be defined in advance.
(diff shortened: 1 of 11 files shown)


DanySK:
How does `ServerState` work?

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/state/ServerState.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import it.unibo.alchemist.common.model.surrogate.EnvironmentSurrogate
+import it.unibo.alchemist.common.model.surrogate.PositionSurrogate
+import it.unibo.alchemist.common.state.CommonState
+import it.unibo.alchemist.core.interfaces.Simulation
+import it.unibo.alchemist.server.state.reducers.environmentSurrogateReducer
+import it.unibo.alchemist.server.state.reducers.simulationReducer
+
+/**
+ * The state of the server.
```
DanySK, line 20: @AngeloFilaseta can you document this class better, linking to the pattern and explaining the rationale behind the singleton state, like you did in the discussion comment?

DanySK (changes requested):

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/state/ServerState.kt
AngeloFilaseta: Perfect, I'll do the same thing for the ClientState.

DanySK: OK. Please resolve this conversation once done, and the PR will automerge.

PR #1851, opened 2023-01-12
feat(loading): allow precision limiting when exporting data. Fixes #1164

### LINKED ISSUE #1164: Double precision should be configurable in exported data
Currently, when we export double data from Alchemist simulations, it does not allow us to configure the precision of the number that will be exported. In this way, the generated CSV, contains unuseful information, producing unnecessary large files.

DanySK:


alchemist-loading/src/main/kotlin/it/unibo/alchemist/loader/m2m/SimulationModel.kt
```diff
@@ -407,12 +422,34 @@ internal object SimulationModel {
...
-                    visitBuilding(context, root)
+                    visitBuilding<Extractor<E>>(context, root)
                 } else {
                     val property = root[DocumentRoot.Export.Data.property]?.toString()
                     val filter: FilteringPolicy = root[DocumentRoot.Export.Data.valueFilter]
                         ?.let { CommonFilters.fromString(it.toString()) }
                         ?: CommonFilters.NOFILTER.filteringPolicy
+                    val precision: Int? = when (val digits = root[DocumentRoot.Export.Data.precision]) {
+                        null -> null
+                        is Int -> digits
```
cric96, line 433: I dunno if it makes sense, but I would prefer something like this:
```
null, is Int -> digits
```
Why you need two separated cases?

cric96 (commented):
just a stylistic doubt

alchemist-loading/src/main/kotlin/it/unibo/alchemist/loader/m2m/SimulationModel.kt
DanySK: Yes, nice, please PR me with that!

PR #1867, opened 2023-01-16
style(alchemist-loader): unify cases in pattern matching at Simulation Model

cric96:


DanySK:
Nice idea, but it cannot compile: they are two different extension methods, and extension methods are resolved statically.

(the JVM equivalent would be `int toInt(byte self)` and `int toInt(short self)`)

DanySK (changes requested):

cric96:
> Nice idea, but it cannot compile: they are two different extension methods, and extension methods are resolved statically.
> 
> (the JVM equivalent would be `int toInt(byte self)` and `int toInt(short self)`)

me

DanySK:
`Type mismatch: inferred type is Any? but Int? was expected`

Dafuq, nice inference Kotlin compiler :laughing:

DanySK:
Closing as Kotlin's type inferrer cannot generalize in this case.

PR #1869, opened 2023-01-17
feat(web-renderer): add Server Routes logic

AngeloFilaseta:
This PR includes the routes logic of the Web Renderer Server.

- **/environment/client**: responds with the Environment in serialized form;
- **/environment/server**: responds with the Environment in rendered form;
- **/simulation/status**: responds with the status of the Simulation;
- **/simulation/pause**: pauses the Simulation;
- **/simulation/play**: plays the Simulation;
- **/**: responds swith the index.html page.

_Note_: web socket could be used in the future, but the logic would remain more or less the same.

### DIFF
--- a/alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
+++ b/alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2010-2023, Danilo Pianini and contributors
+ * listed, for each module, in the respective subproject's build.gradle.kts file.
+ *
+ * This file is part of Alchemist, and is distributed under the terms of the
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.common.utility
+
+/**
+ * Objects that store all the route path. Since this object is part of the common sourceSet, both Client and Serve can
+ * refer to the correct Route maintaining consistency and reducing the number of possible errors and bug.
+ */
+object Ruotes {
+    /**
+     * Base environment path, for all Environment related operations.
+     */
+    private const val environmentPath: String = "/environment"
+
+    /**
+     * Route to get an Environment in a serialized form that needs to be rendered by the client, as
+     * [it.unibo.alchemist.common.model.RenderMode.CLIENT] was requested.
+     */
+    const val environmentClientPath: String = "$environmentPath/client"
+
+    /**
+     * Route to get an Environment already renderer by the Server, as
+     * [it.unibo.alchemist.common.model.RenderMode.SERVER] was requested.
+     */
+    const val environmentServerPath: String = "$environmentPath/server"
+
+    /**
+     * Base simulation path, for simulation related operations.
+     */
+    private const val simulationPath: String = "/simulation"
+
+    /**
+     * Route to get the [it.unibo.alchemist.common.model.surrogate.StatusSurrogate] of the simulation.
+     */
+    const val simulationStatusPath: String = "$simulationPath/status"
+
+    /**
+     * Route to start the simulation.
+     */
+    const val simulationPlayPath: String = "$simulationPath/play"
+
+    /**
+     * Route to pause the simulation.
+     */
+    const val simulationPausePath: String = "$simulationPath/pause"
+}
--- a/alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/modules/InstallModule.kt
+++ b/alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/modules/InstallModule.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2010-2022, Danilo Pianini and contributors
+ * listed, for each module, in the respective subproject's build.gradle.kts file.
+ *
+ * This file is part of Alchemist, and is distributed under the terms of the
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.server.modules
+
+import io.ktor.http.HttpMethod
+import io.ktor.serialization.kotlinx.json.json
+import io.ktor.server.application.Application
+import io.ktor.server.application.install
+import io.ktor.server.plugins.compression.Compression
+import io.ktor.server.plugins.compression.gzip
+import io.ktor.server.plugins.contentnegotiation.ContentNegotiation
+import io.ktor.server.plugins.cors.routing.CORS
+
+/**
+ * Ktor module that adds all the [install] configuration to the application.
+ * The [io.ktor.serialization.kotlinx.json.JsonSupportKt.json] format is used for content negotiation.
+ * A cors configuration is included.
+ * Finally, an encoder for [Compression] is included.
+ * @see <a href="https://ktor.io/docs/modules.html">Ktor Modules</a>
+ */
+fun Application.installModule() {
+    install(ContentNegotiation) {
+        json()
+    }
+    install(CORS) {
+        allowMethod(HttpMethod.Get)
+        allowMethod(HttpMethod.Post)
+        anyHost()
+    }
+    install(Compression) {
+        gzip()
+    }
+}
--- a/alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/EnvironmentRoute.kt
+++ b/alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/EnvironmentRoute.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2010-2022, Danilo Pianini and contributors
+ * listed, for each module, in the respective subproject's build.gradle.kts file.
+ *
+ * This file is part of Alchemist, and is distributed under the terms of the
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.server.routes
+
+import io.ktor.server.routing.Route
+import io.ktor.server.routing.get
+import it.unibo.alchemist.common.model.serialization.encodeEnvironmentSurrogate
+import it.unibo.alchemist.common.model.serialization.jsonFormat
+import it.unibo.alchemist.common.renderer.Bitmap32Serializer
+import it.unibo.alchemist.common.utility.Ruotes.environmentClientPath
+import it.unibo.alchemist.common.utility.Ruotes.environmentServerPath
+import it.unibo.alchemist.server.state.ServerStore.store
+import it.unibo.alchemist.server.utility.Response
+import it.unibo.alchemist.server.utility.Response.Companion.respond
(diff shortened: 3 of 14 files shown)


alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
```diff
@@ -0,0 +1,53 @@
...
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.common.utility
+
+/**
+ * Objects that store all the route path. Since this object is part of the common sourceSet, both Client and Serve can
+ * refer to the correct Route maintaining consistency and reducing the number of possible errors and bug.
+ */
+object Ruotes {
```
DanySK, line 16:
```suggestion
object Routes {
```

alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
```diff
@@ -0,0 +1,53 @@
...
+     */
+    const val environmentClientPath: String = "$environmentPath/client"
+
+    /**
+     * Route to get an Environment already renderer by the Server, as
+     * [it.unibo.alchemist.common.model.RenderMode.SERVER] was requested.
+     */
+    const val environmentServerPath: String = "$environmentPath/server"
+
+    /**
```
DanySK, line 34: Can you explain better what a "simulation related operation" is?

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/modules/InstallModule.kt
```diff
@@ -0,0 +1,40 @@
...
+    install(ContentNegotiation) {
+        json()
+    }
+    install(CORS) {
+        allowMethod(HttpMethod.Get)
+        allowMethod(HttpMethod.Post)
+        anyHost()
+    }
+    install(Compression) {
+        gzip()
```
DanySK, line 38: Do I understand correctly that this means we send around gzip-compressed JSONs? Is there a quick handle for zstd instead?

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/SimulationRoute.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Private function that checks the simulation Status, executes the requested action if possible and responds.
+     * @param action the action to execute on the simulation.
+     * @param additionalCheck a [Pair] representing an additional Status check, first is an invalid [Status] and second
+     * is the error message.
+     */
+    private suspend fun PipelineContext<Unit, ApplicationCall>.respondAction(
+        action: (Simulation<Any, Nothing>) -> Unit,
+        additionalCheck: Pair<Status, String>
```
DanySK, line 92:
```suggestion
        additionalCheck: Pair<Status, String>,
        action: (Simulation<Any, Nothing>) -> Unit,
```

Always make functions the last argument to enjoy the trailing lambda convention

alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Objects that store all the route path. Since this object is part of the common sourceSet, both Client and Serve can
+ * refer to the correct Route maintaining consistency and reducing the number of possible errors and bug.
+ */
+object Ruotes {
+    /**
+     * Base environment path, for all Environment related operations.
+     */
+    private const val environmentPath: String = "/environment"
```
DanySK, line 20: Question: how do we deal with multiple simulations (hence, multiple environments) running on the server?

DanySK (changes requested):
There are a few points I'd like more information about

alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
AngeloFilaseta: So many places I could have gone wrong and I picked the worst 🤦🏽

alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
AngeloFilaseta: Operations that acts directly on the Simulation interface, calling its methods. Is this grouping ok? Do I need to change it or I just document it better?

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/modules/InstallModule.kt
AngeloFilaseta: Yes, that is correct. The [compression algorithms](https://ktor.io/docs/compression.html) included in Ktor are just gzip and deflate. However, I think it may be possible to create a custom Compression Encoder and Decoder just by implementing some interface in the [Compression package](https://api.ktor.io/ktor-server/ktor-server-plugins/ktor-server-compression/io.ktor.server.plugins.compression/-compression.html).
Do I leave it like this for now?

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/EnvironmentRoute.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Route of type GET that retrieve current Environment.
+     * The server will send the environment to the client in a serialized form.
+     * The HTTP [Response] sent to the client can be of type:
+     * 200 (OK) the Environment is sent to the client.
+     */
+    fun Route.environmentClientMode() {
+        get(environmentClientPath) {
+            respond(Response(content = jsonFormat.encodeEnvironmentSurrogate(store.state.environmentSurrogate)))
```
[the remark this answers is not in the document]
AngeloFilaseta: Some [formats](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md) are still experimental. Protobuf can be added as another dependency and uses different APIs. There is not a way to strategy-out the serializer used natively, but it certainly can be implemented by hand. However, I would personally wait until those dependencies are not experimental anymore, especially if the intent is to use them in production.

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/SimulationRoute.kt
AngeloFilaseta: That is correct, will do

alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/common/utility/Ruotes.kt
AngeloFilaseta: **_alchemist-web-renderer_** takes one configuration file and run the simulation while also offering a web server for accessing its state. The easiest way that comes to my mind is just to launch _n_ instances of _web-renderer_ on the server. Every instance would take a different port and work as expected. In this case the limitation would be on the front-end, as two different tab are needed on the browser.

DanySK: What happens if I launch a batch?
Alchemist configuration files do not describe a single simulation, but a batchable set of them

AngeloFilaseta: Batch mode is still not supported. It would still be possible to implement it, but the ServerState and the Routes Logic should be modified a bit . Finally a Button to switch between the Simulations on the front-end should be added.
With the current architecture this change should not be that hard to achieve, but it is still not present.
I tried inspiration from the Swing Gui and the Fx gui, since this project is similar from certain points, and as I see (correct me if I'm wrong) none of them supports the batch mode.
Anyway I'm keeping tracks of all the features that are still missing from the beginning of the project and how they could be achieved by changing the existent codebase. If you think that's okay I would add them to the "Future work" section of my thesis, so anyone who would work on this project after me would have a easier life.

DanySK: @AngeloFilaseta ok. The difference with the other two UIs is that this one is distributed, and the server is meant to run on a capable machine, with users on the network that may want to run simulations on it concurrently. Add this feature to the top of the future work, as soon as we have the current state merged and the thesis mostly written, if there is time I'd like to tackle this issue.

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/EnvironmentRoute.kt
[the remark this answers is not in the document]
DanySK: I see your point, but the concept of "experimental" in Kotlin is more stable than the concept of "stable" in Alchemist, being the latter a simulator for scientific research :wink: 

So in kotlinx.serialization every serializer exposes a different API?

DanySK (approved):
only a minor clarification

alchemist-web-renderer/src/jvmMain/kotlin/it/unibo/alchemist/server/routes/EnvironmentRoute.kt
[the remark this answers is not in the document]
AngeloFilaseta: Yes, please check these links with some examples:
[Protobuf API example](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#protobuf-experimental)
[CBOR API example](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#cbor-experimental)
As you can see the Classes used here differs, same thing happens for Json. There's one thing I didn't mention though, don't know why I didn't say it first. Remember this?
```kotlin
install(ContentNegotiation) {
    json()
}
```
By using this code in the Ktor install module we instruct the routes on the server to automatically use serializers if the object that we want to send is an instance of a serializable class. Same thing happens on the client side for deserialization. In other words, I could have done a simple:
```kotlin
fun Route.environmentClientMode() {
  get(environmentClientPath) {
      respond(Response(store.state.environmentSurrogate))
  }
}
 ```
There's a problem though. The ``EnvironmentSurrogate`` class needs a specific Polymorphic serializer because of the two type parameters. Ktor still does not support Polymorphic serializers in content negotiation, so I had to manually create that annoying ``encodeEnvironmentSurrogate`` extension function.
This was the bad news. 
The good news is: changing that ``json()`` to ``protobuf()`` would be enough to accomplish what you want, but only if this limit is surpassed in some way. For now, the only valid open option that comes to my mind is to create a function that would manually check the Content Negotiation strategy being used (hoping this is possible) and change the serialization API accordingly.

Also note that the API changes, but the structure of the Surrogate Classes would remain the same.

AngeloFilaseta:
@DanySK do I need to change anything? I can resolve all the conversations if everything is good and proceed :)

DanySK:
If they are solved, yes, please, do.

PR #1871, opened 2023-01-17
build(web-renderer): simplify shadowJar task configuration

AngeloFilaseta:
This PR is intended to simplify the alchemist-web-renderer build.

- The **shadowJar** task has been configured using a different strategy than before. The **jvmJar** task output (from Kotlin Multiplatform) is now used to generate a fat jar that includes the JVM classes. The **jsBrowserProductionWebpack** output is also included in the fat jar, like it was before these changes.
**_Note_**: The problem encountered in #1709 COULD be resolved with these changes. I believe the problem is caused by the use of the `finalizedBy` method. In this implementation, that use was completely removed.

- The run task now depends explicitly on the **shadowJar** task. This will ensure that the **shadowJar** task runs before the **run** task and that the output file of the **shadowJar** task is added to the classpath when the **run** task is executed.

### DIFF
--- a/alchemist-web-renderer/build.gradle.kts
+++ b/alchemist-web-renderer/build.gradle.kts
@@ -89,7 +89,9 @@ application {
 }
 
 tasks.getByName<JavaExec>("run") {
-    classpath(tasks.getByName<Jar>("shadowJar"))
+    val shadowJarTask = tasks.getByName<Jar>("shadowJar")
+    dependsOn(shadowJarTask)
+    classpath(shadowJarTask)
 }
 
 /**
--- a/alchemist-web-renderer/build.gradle.kts
+++ b/alchemist-web-renderer/build.gradle.kts
@@ -11,7 +11,6 @@ import Libs.alchemist
 import Libs.incarnation
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import org.jetbrains.kotlin.gradle.targets.js.webpack.KotlinWebpack
-import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
 
 plugins {
     application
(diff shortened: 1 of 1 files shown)


alchemist-web-renderer/build.gradle.kts
```diff
@@ -90,35 +89,24 @@ application {
 }
 
 tasks.getByName<JavaExec>("run") {
-    classpath(tasks.getByName<Jar>("shadowJar"))
+    val shadowJarTask = tasks.getByName<Jar>("shadowJar")
+    dependsOn(shadowJarTask)
```
DanySK, line 93: This seems to imply that I need to pack a fatJar in order to run the application -- why so?

alchemist-web-renderer/build.gradle.kts
```diff
@@ -90,35 +89,24 @@ application {
 }
 
 tasks.getByName<JavaExec>("run") {
```
DanySK, line 91: Does a plain:

```suggestion
tasks.run {
```

work?

In any case, avoid `getByName`, use the [configuration avoidance API](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html). I usually do not enforce it, but the build of Alchemist is already too slow to leave things sub-optimal.

alchemist-web-renderer/build.gradle.kts
```diff
@@ -90,35 +89,24 @@ application {
...
-    classpath(tasks.getByName<Jar>("shadowJar"))
+    val shadowJarTask = tasks.getByName<Jar>("shadowJar")
```
DanySK, line 92:
```suggestion
    val shadowJarTask = tasks.shadowJar
```

shoud work as well

alchemist-web-renderer/build.gradle.kts
```diff
@@ -90,35 +89,24 @@ application {
...
-                archiveClassifier.set("all")
-            }
-            listOf("jvmJar", "shadowJar").forEach {
-                tasks.named<Jar>(it) {
-                    finalizedBy(shadowJar)
-                }
-            }
-        }
-    }
+tasks.withType<ShadowJar> {
```
DanySK, line 101: configuration avoidance
```suggestion
tasks.withType<ShadowJar>().configureEach {
```

alchemist-web-renderer/build.gradle.kts
```diff
@@ -90,35 +89,24 @@ application {
...
+    val jvmJarTask = tasks.getByName<Jar>("jvmJar")
```
DanySK, line 102: either
```suggestion
    val jvmJarTask = tasks.jvmJar
```
or
```suggestion
    val jvmJarTask = tasks.named("jvmJar")
```

alchemist-web-renderer/build.gradle.kts
```diff
@@ -90,35 +89,24 @@ application {
...
+    val webpackTask = tasks.getByName<KotlinWebpack>("jsBrowserProductionWebpack")
```
DanySK, line 103: Same as above

DanySK (changes requested):
Stuff mostly related to configuration avoidance and reduced use of string/type-unsafe access

DanySK:
@AngeloFilaseta you let me know when it's ready for the next review round.

alchemist-web-renderer/build.gradle.kts
AngeloFilaseta: The [Kotlin Multiplatform Full Stack App example](https://kotlinlang.org/docs/multiplatform-full-stack-app.html) uses this trick to have acces to all the compiled files and resources, even the javascript one, simplifiyng the build. I can try to use the output of compile tasks though, that should work too and I think I can do it. I'm gonna update this one I found out.

alchemist-web-renderer/build.gradle.kts
AngeloFilaseta: Ok, I'll remove this

alchemist-web-renderer/build.gradle.kts
DanySK: I cannot find where. Looks like the opposite from that page, the `jvmJar` task depends on a `webpackTask`.
Here we are saying that, in order to run, we need to pack a fatJar (and test it, since there is a finalizer).

Or do you refer to:
```kotlin
tasks.getByName<JavaExec>("run") {
    classpath(tasks.getByName<Jar>("jvmJar")) // so that the JS artifacts generated by `jvmJar` can be found and served
}
```

?

AngeloFilaseta:
@DanySK I updated the build by following the Configuration Avoidance guide. I think it is much better now. I closed most of the outdated conversations. Check especially the code that I pointed out in the conversation that is still open to see if the change is ok or not. Thank you!

PR #1873, opened 2023-01-18
docs(website): make the landing page nicer and more useful

DanySK:
New aspect:

image


DanySK:
@mviroli if you have additional suggestions before this PR gets rebased into the mainline, they are welcome :)

site/content/_index.md
```diff
@@ -7,12 +7,28 @@ weight = 5
...
 # ![logo](images/logo-text-path.svg)
 
-This website hosts the documentation of the Alchemist Simulator.
-The simulator, created towards the end of 2010 as a by-product of the European Project SAPERE,
-has since grown, and it is now a full-fledged simulation framework.
+This website hosts the documentation of the Alchemist Simulator,
+created towards the end of 2010 as a by-product of the European Project SAPERE,
+and grown ever since into a full-fledged simulation framework.
+
+<video loop autoplay style="max-width: 100%; display: inline-block; ">
```
cric96, line 14: I think it is better to use a shortcode here. We could leverage [hugo-video](https://github.com/martignoni/hugo-video)

DanySK: Nice, but do you think it is worth having yet another submodule just for that? Consider that, in most cases, we embed already with the youtube shortcode...

Issue #1939, opened 2023-02-07
Unable to distribute simulation batches


kelvin-olaiya:
I had a look at the current state of the `alchemist-grid` module. 
I've problems in trying to distribute batches on a single node cluster (aka, the cluster is my laptop).

What i did was:

1. Start Alchemist in server mode. After reading the [Apache Ignite documentation](https://ignite.apache.org/docs/latest/understanding-configuration), I understand that i need to provide a configuration file with Alchemist `-s` CLI option. This is the configuration file i used:

```xml
<?xml version="1.0" encoding="UTF-8"?>

<beans xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="
        http://www.springframework.org/schema/beans
        http://www.springframework.org/schema/beans/spring-beans.xsd">

    <bean class="org.apache.ignite.configuration.IgniteConfiguration">
        <property name="workDirectory" value="/home/kelvin/Development/ignite"/>
    </bean>
</beans>
```
2. Launch Alchemist in distributed batch mode, providing the same configuration file to the `-d` cli option.

Trying doing the latter results in errors that seem to have to do with serialization. [Here](https://github.com/AlchemistSimulator/Alchemist/files/10673882/log.txt) you can see the error log.

DanySK:
Apparently an object is not serializable: how does Ignite serialize?

DanySK:
Can you deepen the reason why it does refuse to serialize? It looks like it's failing at serializing Kotlin's lambda expressions - maybe we can program it?

kelvin-olaiya:
Ok, I'll try to debug

kelvin-olaiya:
Most of the serialization exceptions seem to origin from classes whose fields in some ways end up dealing with `org.danilopianini.jirf.FunctionEdge`. I guess the problem may be that it does not implement Serializable.

Other classes with similar issue are:
* `org.danilopianini.jirf.FactoryImpl`
* `it.unibo.alchemist.loader.m2m.Context`
* `org.jgrapht.graph.DefaultDirectedGraph` (although this one extends a class who does implement Serializable) 
* `it.unibo.alchemist.loader.m2m.SimulationModel$fromMap$3`

And then there's another kind of exception regarding which I could not find so many information.

```
[...]

PR #1964, opened 2023-02-13
refactor(alchemist-grid): deal with serialization issues 

kelvin-olaiya:
I was finally able to "distribute" a simulation batch from a client to an Alchemist server. The procedures are the one summarized in #1939. 
@DanySK this PR is to show what I did in order to solve some previous serialization problems and get some feedbacks, so there's no intention to merge.


alchemist-grid/src/main/java/it/unibo/alchemist/grid/config/LightInfoGeneralSimulationConfig.java
```diff
@@ -25,7 +26,7 @@ public abstract class LightInfoGeneralSimulationConfig implements GeneralSimulat
     private static final long serialVersionUID = 1L;
     private final long endStep;
     private final Time endTime;
-    private final Loader loader;
+    private transient final Loader loader;
```
DanySK, line 29: What happens on deserialization?

alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/terminators/StableForSteps.kt
```diff
@@ -47,6 +47,7 @@ class StableForSteps<T>(
 ) : Predicate<Environment<T, *>> {
     private var success: Long = 0
     private var positions: Map<Node<T>, Position<*>> = emptyMap()
+    @Transient
     private var contents = makeTable<T>(0)
```
DanySK, line 51: what happens on deserialization?

alchemist-grid/src/main/java/it/unibo/alchemist/grid/config/LightInfoGeneralSimulationConfig.java
kelvin-olaiya: The problem was in serialization, some of the loader contents do not implement Serializable (`org.danilopianini.jirf.FactoryImpl`, `org.jgrapht.graph.DefaultDirectedGraph`, etc.)

alchemist-implementationbase/src/main/kotlin/it/unibo/alchemist/model/implementations/terminators/StableForSteps.kt
kelvin-olaiya: I'm not quite sure, the problem was a ClassNotFoundException:
```
it.unibo.alchemist.model.implementations.terminators.StableForSteps$Companion$$Lambda$1058/0x0000000800885440
```
what I supposed was that being the class Initialized locally and then distributed, the remote node can't find the "instance" of the lambda function in the `makeTable` function.

DanySK: I mean: that "contents" goes to null upon deserialization. It is at a risk of throwing NPE

kelvin-olaiya: "contents" is only used in equality check, and eventually get overwritten. 
The first time it may compare with null instead of an empty table and then with the table with the actual contents. If a user wants to terminate after 1 stable step, it won't be accurate.

DanySK:
Is this one going to be merged and fixed, or should I close it?

kelvin-olaiya:
This one can be closed. I'll eventually open a new one with An alternative solution.

DanySK:
Okay, thanks

PR #2023, opened 2023-02-21
fix(alchemist-engine): fix #2017

### LINKED ISSUE #2017: `goToTime` should work multiple times
The `goToTime` method in the `Simulation` should allow users to move the simulation to a specific point in the future and pause it. However, after testing this method, I discovered that there is a flaw in its implementation.

When the `goToTime` method is first called, the `Simulation` correctly jumps to the specified time using the `DoubleTime` parameter:
```
val simulation = ...
simulation.goToTime(DoubleTime(10.0)
simulation.play()
// run simulation into another thread
executor. submit { simulation. run() }
simulation.waitFor(Stautus.PAUSED, 1, TimeUnit.SECONDS)
// here simulation.getTime is greater than 10.0
```

The issue arises when the `goToTime` method is called a second time to jump to a new time, as the simulation does not progress past the previous paused time. This problem occurs because the `OutputMonitor` is not removed after the first time the simulation is paused, which causes the monitor to trigger again and pause the simulation at the previous time.

Upon examining the source code, it is clear that the issue is caused by the OutputMonitor not being removed after its initial use. I have created a test that demonstrates this error, which can be found [here](https://github.com/AlchemistSimulator/Alchemist/blob/0fe462fcb9cd0f96c66b6eee8ad1449b378f7bae/alchemist-engine/src/test/kotlin/it/unibo/alchemist/core/test/TestSimulationControl.kt).

cric96:


DanySK:
Approved - if tests pass, of course :)

alchemist-engine/src/test/kotlin/it/unibo/alchemist/core/test/TestSimulationControl.kt
```diff
@@ -51,7 +53,7 @@ class TestSimulationControl : FreeSpec(
                     time shouldBe firstJump
                     goToTime(nextJump)
                     play()
-                    workerPool.submit { run() }
+                    delay(awaitForNextJump) // best solution so far for now, you cannot be sure that the simulation is played again
```
cric96, line 56: @DanySK I know that this is really fragile -- I can try with something more sophisticated, but at least for the test should be ok, what do you think?

DanySK: Merging but taking your proposal for a fresh API as a promise :smile:

PR #2032, opened 2023-02-24
feat(alchemist-api): add async simulation control

cric96:
@DanySK we need to discuss about some choices:
- we wanna maintain waitFor? Or we can cover all the uses cases leveraging `CompletableFuture`?
- `goToX` now return a `CompletableFuture` that will be completed when the simulation eventually reach that X (time or step). It makes sense?
- there are other methods that should be async in your opinion?
I create the PR in draft because I need to create some test suite for the new API, but I want to start the discussion in order to understand if we wanna introduce some other changes.


DanySK:
* It depends on whether the same case is captured otherwise. It seems to me that it is captured by `goToTime`

I think that there is a general issue of "reacting to an event". Say that I want to do something when the simulator is paused. I send a pause command and wait on the completable future. Still, I have no guarantee that by the time the future is populated, the simulator is still paused.

We might want to have an additional optional callback, something like:

```kotlin
@JvmOverloads
fun pause(andThen: Simulation<T>.() -> Unit = { }): CompletableFuture<Simulation<T>>
```

Or can we capture the "as soon as it's paused" already?

Also, one may want to do something as soon as some condition applies. There are several approaches, the simplest is to take a boolean that tells you whether to reschedule. Otherwise, taking two lambdas, or return an object that allows to be re-called with the scheduled action.

```kotlin
@JvmOverloads
fun schedule(operation: Simulation<T>.() -> Boolean = { true }): Unit
```

to be used like:

```kotlin
simulation.schedule {
    mycondition().onTrue { mySideEffect() }
}
```

cric96:
> I think that there is a general issue of "reacting to an event". Say that I want to do something when the simulator is paused. I send a pause command and wait on the completable future. Still, I have no guarantee that by the time the future is populated, the simulator is still paused.

I understand your point. With `CompletableFuture`, the semantics are such that the callback will only run once the event has succeeded at least once. While there is a method to run a callback in the same thread as the CompletableFuture's completion, there is no guarantee that the callback will be called immediately after the future is completed. Here's an example to illustrate this:

```
val future = new CompletableFuture[Int]()
val completeThread = new Thread(() => {
  future.complete(10)
  println("After completed")
})
completeThread.start()
future.thenRun(() => println("Future completed"))
```
In this case, it's possible that "After completed" will be printed before "Future completed", since the future may complete before the `println` statement in the thread that was started by completeThread.
> We might want to have an additional optional callback, something like:
> 
> ```kotlin
> @JvmOverloads
> fun pause(andThen: Simulation<T>.() -> Unit = { }): CompletableFuture<Simulation<T>>
> ```
> 
> Or can we capture the "as soon as it's paused" already?

I replay above... 
BTW I believe it's still preferable to avoid passing the `andThen` callback through the pause method, as this would cause us to lose the composability of the future.

> 
> Also, one may want to do something as soon as some condition applies. There are several approaches, the simplest is to take a boolean that tells you whether to reschedule. Otherwise, taking two lambdas, or return an object that allows to be re-called with the scheduled action.
> 
> ```kotlin
> @JvmOverloads
> fun schedule(operation: Simulation<T>.() -> Boolean = { true }): Unit
> ```
> 
> to be used like:
> 
> ```kotlin
> simulation.schedule {
>     mycondition().onTrue { mySideEffect() }
> }
> ```
Is it common to create APIs like this in Kotlin? Personally, I'm not a fan of "hiding" side effects in a predicate, but I suppose it's a common practice in languages without currying. In Scala, for example, I would write something like:
```
def schedule(condition: => Boolean)(operation: => Unit): Unit
```
This allows you to schedule the operation only if the condition is true (that makes more sense for me...)

DanySK:
> > I think that there is a general issue of "reacting to an event". Say that I want to do something when the simulator is paused. I send a pause command and wait on the completable future. Still, I have no guarantee that by the time the future is populated, the simulator is still paused.
> 
> I understand your point. With `CompletableFuture`, the semantics are such that the callback will only run once the event has succeeded at least once. While there is a method to run a callback in the same thread as the CompletableFuture's completion, there is no guarantee that the callback will be called immediately after the future is completed. Here's an example to illustrate this:
> 
> ```
> val future = new CompletableFuture[Int]()
> val completeThread = new Thread(() => {
>   future.complete(10)
>   println("After completed")
> })
> completeThread.start()
> future.thenRun(() => println("Future completed"))
> ```
> 
> In this case, it's possible that "After completed" will be printed before "Future completed", since the future may complete before the `println` statement in the thread that was started by completeThread.
> 
> > We might want to have an additional optional callback, something like:
> > ```kotlin
> > @JvmOverloads
> > fun pause(andThen: Simulation<T>.() -> Unit = { }): CompletableFuture<Simulation<T>>
> > ```

Yes. My understanding is that with a callback API this won't be an issue - and the `CompletableFuture` API can stand on top (as using CPS gets shitty pretty soon).
I don't know if Kotlin's coroutines can be an alternative here.

> I replay above... BTW I believe it's still preferable to avoid passing the `andThen` callback through the pause method, as this would cause us to lose the composability of the future.

I don't understand why. Do you have an example?

> > Also, one may want to do something as soon as some condition applies. There are several approaches, the simplest is to take a boolean that tells you whether to reschedule. Otherwise, taking two lambdas, or return an object that allows to be re-called with the scheduled action.
> > ```kotlin
> > @JvmOverloads
> > fun schedule(operation: Simulation<T>.() -> Boolean = { true }): Unit
> > ```
> >   
> > to be used like:
> > ```kotlin
> > simulation.schedule {
> >     mycondition().onTrue { mySideEffect() }
> > }
> > ```
> 
> Is it common to create APIs like this in Kotlin? Personally, I'm not a fan of "hiding" side effects in a predicate, but I suppose it's a common practice in languages without currying. In Scala, for example, I would write something like:
> 
> ```
> def schedule(condition: => Boolean)(operation: => Unit): Unit
> ```

The point with Groovy and Kotlin is that calls get composed in blocks by using the trailing lambda convention, passing two functions to a method often makes the API much less idiomatic at the caller site.

Anyway, I see your point. The first way that comes to my mind to emulate currying is to do something like:

```kotlin
object ComputeSlot {
    fun do(operation: Simulation<T>.() -> Unit)
}
fun schedule(operation: Simulation<T>.() -> Unit): Unit = asSoonAs { true }.do(operation)
fun asSoonAs(condition: Simulation<T>.() -> Boolean): ComputeSlot = TODO()
```

One could override the `operator fun invoke` in `ComputeSlot` and get the Scala-like `asSoonAs { true }{ TODO() }`, but I'm not sure the Kotlin compiler would be happy (also, it is much less clear for a Kotlin user).

(note: not tested, just to provide an idea)

> This allows you to schedule the operation only if the condition is true (that makes more sense for me...)

I think that the effect is exactly the same. The operation is scheduled only if the condition is true. Maybe we should have a set of "interactions we'd like to have with the simulator" in Kotlin and Scala, and figure out an API to support them all.

DanySK (approved):

cric96:
We said that we would wait to do this in the new responsive engine. 
But btw sure I can finalize it. I'll see if I can tonight :) 
@DanySK

DanySK:
Meh, let's patch this and then if we ever can have a reactive engine we'll do things better

DanySK:
@cric96 I don't get why the CI doesn't trigger on this one... Any hint?

PR #2034, opened 2023-02-24
fix(ci): add uploadJvm and uploadJs to publishCmd

AngeloFilaseta:
I believe we forgot to integrate the changes in dce26ef917173277dd6c588713e94954bc6f3b41.
With this PR the correct artifacts should be correctly uploaded for Multiplatform project on releases.


DanySK:
Merging as `uploadJs` is already in dry-deploy and the mainline passes

PR #2218, opened 2023-04-23
feat: add draw directed node effect

cric96:
Effect used in Macro Swarm paper: https://github.com/AggregateComputing/experiment-2023-coordination-swarm-behaviour

### DIFF
--- a/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/boundary/swingui/effect/impl/DrawDirectedNode.kt
+++ b/alchemist-swingui/src/main/kotlin/it/unibo/alchemist/boundary/swingui/effect/impl/DrawDirectedNode.kt
@@ -9,7 +9,6 @@
 
 package it.unibo.alchemist.boundary.swingui.effect.impl
 
-import it.unibo.alchemist.boundary.swingui.effect.api.Effect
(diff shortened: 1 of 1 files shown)


DanySK:
@cric96 https://scans.gradle.com/s/nmmb343pwqku6/console-log?task=:alchemist-swingui:compileKotlin

cric96:
> @cric96 https://scans.gradle.com/s/nmmb343pwqku6/console-log?task=:alchemist-swingui:compileKotlin

Do I need to add @file:Suppress("DEPRECATION") for the deprecation warning since the effect is deprecated, right? Also, I'm not very familiar with the Kotlin type system, so I was wondering if you have any tips on how to handle the unsafe cast situation. Your insights would be greatly appreciated! :))

DanySK:
it's better to suppress for the class rather than for the entire file:

@Suppress("DEPRECATION")

DanySK:
Type system wise, the Kotlin type system is in between Java and Scala (and closer to Java). No path-dependent types, no HKT, in this case you can't do much, you can suppress pointwise: `@Suppress("UNCHECKED_CAST")`.

I've sent you a commit with some changes to improve idiomaticity. Note: there are still style errors.

PR #2352, opened 2023-06-06
feat(alchemist-loading): include output monitor in the configuration file

cric96:


cric96:
Ok! I should have done 1. and 2. 
For point 3. : do you have some good examples that we can include? By now I just put a really simple use case.

DanySK:
@cric96 a simple use case is okay. Can you sync this with the mainline?

cric96:
@DanySK I believe there's an issue with the action designated for verifying the presence of links. While the path exists locally, it appears to be checking the site that's currently deployed, rather than the one being built. Is my understanding correct?

PR #2495, opened 2023-07-29
feat!: new command line

KNtronov:
- Re-implement CLI in kotlinx-cli
- Move launch parameters into configuration
- Implement config file override in CLI


alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/LoadAlchemist.kt
```diff
@@ -28,47 +29,106 @@ object LoadAlchemist {
...
+        SimulationModel.fromMap(
+            applyOverrides(model.from(reader), overrides),
+        )
 
     /**
      * Load from a [reader].
      */
     @JvmStatic
-    fun from(reader: Reader, model: AlchemistModelProvider) = SimulationModel.fromMap(model.from(reader))
+    fun from(reader: Reader, model: AlchemistModelProvider) =
```
KNtronov, line 61: An alternative and more idiomatic kotlin approach would be to have a default value for overrides in the signature:

,overrides: List<String> = emptyList()

This however causes some issues when inter-oping with Java (when calling kotlin from java) because java doesn't automatically infer the default value and requires it to be passed.

Since we have some java tests calling into this class, I preferred this more verbose solution, but safer.

alchemist-loading/src/main/kotlin/it/unibo/alchemist/model/util/VariablesOverrider.kt
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (C) 2010-2023, Danilo Pianini and contributors
+ * listed, for each module, in the respective subproject's build.gradle.kts file.
+ *
+ * This file is part of Alchemist, and is distributed under the terms of the
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.model.util
```
DanySK, line 10:
```suggestion
package it.unibo.alchemist.boundary.loader
```

alchemist-multivesta-adapter/src/main/kotlin/it/unibo/alchemist/multivesta/adapter/AlchemistMultiVesta.kt
```diff
@@ -47,7 +50,6 @@ object AlchemistMultiVesta {
     private const val MULTIVESTA = "mv"
     private const val ALREADY_EXECUTED_SIMULATION = "aes"
     private const val HEADLESS = "hl"
-    private const val VARIABLES = "var"
     private const val BATCH = 'b'
     private const val FXUI = "fxui"
```
DanySK, line 54: This could be tropped, right?

DanySK (changes requested):
Reviewed 15/29. Only minor changes requested

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/LoadAlchemist.kt
KNtronov: This is a much better solution, I've updated the signatures

alchemist-loading/src/main/kotlin/it/unibo/alchemist/model/util/VariablesOverrider.kt
KNtronov: Done

alchemist-loading/src/test/kotlin/it/unibo/alchemist/test/TestVariablesOverrider.kt
```diff
@@ -0,0 +1,71 @@
...
+                        "key1" to "testtest",
+                        "key2" to mapOf(
+                            "foo" to listOf(
+                                10,
+                            ),
+                        ),
+                    ),
+                ),
+            ),
+        )
```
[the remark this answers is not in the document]
KNtronov: Updated the test

alchemist-multivesta-adapter/src/main/kotlin/it/unibo/alchemist/multivesta/adapter/AlchemistMultiVesta.kt
KNtronov: This is a mistake, I have reverted the changes and put the variables back in. Please re-review this file.

alchemist-multivesta-adapter/src/main/kotlin/it/unibo/alchemist/multivesta/adapter/AlchemistMultiVesta.kt
```diff
@@ -77,6 +79,7 @@ object AlchemistMultiVesta {
             when (val value = converter(it)) {
                 null ->
                     exitBecause("Not a valid ${T::class.simpleName}: $it", ExitStatus.NUMBER_FORMAT_ERROR)
+
                 else -> value
```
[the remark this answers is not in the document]
KNtronov: Done

alchemist-web-renderer/src/jvmTest/resources/simulation.yml
```diff
@@ -0,0 +1,17 @@
...
+    end-time: 50
+    is-web: true
+    is-batch: true
+    distributed-configuration-path: foo-bar
+    graphics-path: foo-bar
+    server-configuration-path: foo-bar
+    engine-mode: batchEpsilon
+    epsilon: 0.01
+    batch-size: 10
+    output-replay-strategy: replay
```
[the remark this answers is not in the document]
KNtronov: Done

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package it.unibo.alchemist.boundary.loader
+
+import it.unibo.alchemist.boundary.modelproviders.YamlProvider
+
+/**
+ * Use this to override a map of variables with a list of resolvable key-value pairs.
+ */
+object VariablesOverrider {
```
DanySK, line 17: Probably better to see this 

```suggestion
internal object Overrides {
```

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
```diff
@@ -0,0 +1,62 @@
...
+    @JvmStatic
+    fun applyOverrides(map: Map<String, *>, overrides: List<String>): Map<String, *> {
+        return if (overrides.isEmpty()) {
+            map
+        } else {
+            val newMap = map.toMutableMap()
+            overrides.forEach { applyOverride(it, newMap) }
+            newMap
+        }
+    }
```
DanySK, line 33:
```suggestion
    @JvmStatic
    fun Map<String, *>.overrideAll(overrides: List<String>): Map<String, *> = when {
        overrides.isEmpty() -> this
        else -> 
            this.toMutableMap().also { overrides.forEach { applyOverride(it, newMap) } }
    }
```

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @Suppress("UNCHECKED_CAST")
+    @JvmStatic
+    private fun mergeInto(key: String, value: Any?, newMap: MutableMap<String, Any?>) {
+        when (value) {
+            is MutableMap<*, *> -> {
+                if (newMap[key] is MutableMap<*, *>) {
+                    val castValue = value as MutableMap<String, Any?>
```
DanySK, line 49: this is unchecked... and the name is not very clear. "Castvalue"? maybe "innerMap"?

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
```diff
@@ -0,0 +1,62 @@
...
+                    val pointer = newMap[key] as MutableMap<String, Any?>
```
DanySK, line 50: pointer to what?

alchemist-loading/src/test/kotlin/it/unibo/alchemist/test/TestVariablesOverrider.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import io.kotest.core.spec.style.StringSpec
+import io.kotest.matchers.shouldBe
+import it.unibo.alchemist.boundary.loader.VariablesOverrider
+import it.unibo.alchemist.boundary.modelproviders.YamlProvider
+import org.kaikikm.threadresloader.ResourceLoader
+
+class TestVariablesOverrider : StringSpec({
+
+    "test overrides" {
```
DanySK, line 20:
```suggestion
    "overriding variables in a YAML file should be possible" {
```

alchemist-multivesta-adapter/src/main/kotlin/it/unibo/alchemist/multivesta/adapter/AlchemistMultiVesta.kt
```diff
@@ -212,6 +210,7 @@ object AlchemistMultiVesta {
                     "Conflicting verbosity specification. Only one of ${logLevels.keys} can be specified.",
                     ExitStatus.MULTIPLE_VERBOSITY,
                 )
+
```
DanySK, line 213:
```suggestion
```

src/main/kotlin/it/unibo/alchemist/AlchemistExecutionOptions.kt
```diff
@@ -38,12 +47,20 @@ data class AlchemistExecutionOptions(
     val server: String? = null,
     val parallelism: Int = defaultParallelism,
     val endTime: Double = defaultEndTime,
```
DanySK, line 49: We can drop this. `terminators` are available in the configuration file.

src/main/kotlin/it/unibo/alchemist/AlchemistExecutionOptions.kt
```diff
@@ -54,8 +71,42 @@ data class AlchemistExecutionOptions(
...
 
         private val NO_OPTION = AlchemistExecutionOptions()
+
+        /**
+         * @property engineMode engine execution mode.
+         * @property outputReplayStrategy batch mode engine output replay strategy.
+         * @property batchSize fixed batch mode batch size.
+         * @property epsilon epsilon batch mode epsilon threshold.
+         */
+        data class EngineConfig(
```
DanySK, line 90: This looks like it has percolated from the implementation of the engine. It should not be there, defaults for the engine should be specified in the engine itself

src/main/kotlin/it/unibo/alchemist/config/SimulationConfig.kt
```diff
@@ -0,0 +1,198 @@
...
+     * @property serverConfigPath if launched as Alchemist grid node server,
+     * the path to the configuration file. Null otherwise.
+     * @property engineMode engine execution mode
+     * @property outputReplayStrategy batch engine mode output replay strategy
+     * @property batchSize batch engine mode fixed batch size
+     * @property epsilon epsilon engine mode epsilon threshold value
+     */
+    data class SimulationConfigParameters(
+        val variables: List<String>,
+        val parallelism: Int,
```
DanySK, line 55: depends on the launcher

src/main/kotlin/it/unibo/alchemist/config/SimulationConfig.kt
```diff
@@ -0,0 +1,198 @@
...
+        val endTime: Double,
+        val isWeb: Boolean,
+        val isBatch: Boolean,
```
DanySK, line 58: depends on the launcher

DanySK (changes requested):
There are several major changes needed. It looks like there is a lot more hardcoded stuff than expected.
Engines should be loaded as everything else with type/parameters, thus most parameters / enums do not make sense.

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: I agree, innerMap is much better. As for the unchecked cast, unfortunately I couldn't get around the type erasure limitations. I could not check for 

newMap[key] is MutableMap<String, *>

 as the type gets erased (I get this error at compile time): Cannot check for instance of erased type: MutableMap<String, *>. Same for the "when" block on the value.

I feel like unchecked casts to map with string keys is a neccessary evil here. I agree that this implementations make very strong assumptions, however a map with a non-string value wouldn't make sense in this context since yaml maps are string-keyed so I think that it's acceptable to fail and throw an exception here. Maybe the exception could be more explicit, something like IllegalArgumentException("cannot ovveride a non string-keyed map")?

Is there a better way to check the collection types here that I'm missing?

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: This algorithm basically recursively navigates the tree (if we imagine a map of maps as a tree) and checks: if it's a leaf the value gets overridden, if it's an inner node it continues the navigation. The "pointer" is to the current node to be checked.

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: Done

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: Done

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: Refactored to "innerMap"

alchemist-loading/src/test/kotlin/it/unibo/alchemist/test/TestVariablesOverrider.kt
KNtronov: Done

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
DanySK: Yes, cast should be done to MutableMap<*, *>, and key/values checked and cast upon usage

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
DanySK: so maybe a better name would be something like currentTreeNode?

src/main/kotlin/it/unibo/alchemist/config/SimulationConfig.kt
KNtronov: Removed, see general changes to Loader and Launcher creation

src/main/kotlin/it/unibo/alchemist/config/SimulationConfig.kt
KNtronov: Removed, see general changes to Loader and Launcher creation

src/main/kotlin/it/unibo/alchemist/config/SimulationConfig.kt
```diff
@@ -0,0 +1,198 @@
...
+     * the path to the configuration file. Null otherwise.
+     * @property engineMode engine execution mode
+     * @property outputReplayStrategy batch engine mode output replay strategy
+     * @property batchSize batch engine mode fixed batch size
+     * @property epsilon epsilon engine mode epsilon threshold value
+     */
+    data class SimulationConfigParameters(
+        val variables: List<String>,
+        val parallelism: Int,
+        val endTime: Double,
```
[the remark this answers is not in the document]
KNtronov: Removed, see general changes to Loader and Launcher creation

src/main/kotlin/it/unibo/alchemist/config/OutputReplayStrategy.kt
```diff
@@ -0,0 +1,39 @@
...
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.config
+
+/**
+ * Engine output replay strategy modes.
+ * @property code mode code
+ */
+enum class OutputReplayStrategy(val code: String) {
```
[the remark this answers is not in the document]
KNtronov: Removed, see general changes to Loader and Launcher creation

src/main/kotlin/it/unibo/alchemist/AlchemistExecutionOptions.kt
KNtronov: Removed, see general changes to Loader and Launcher creation. End time is dropped, only terminators are used.

src/main/kotlin/it/unibo/alchemist/config/EngineMode.kt
```diff
@@ -0,0 +1,44 @@
...
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.config
+
+/**
+ * Engine modes.
+ * @property code mode code
+ */
+enum class EngineMode(val code: String) {
```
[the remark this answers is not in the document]
KNtronov: Removed, see general changes to Loader and Launcher creation.

Engine config will be implemented with a similar mechanism in the PR for batch engine as currently it would do nothing.

src/main/kotlin/it/unibo/alchemist/AlchemistExecutionOptions.kt
KNtronov: Removed, see general changes to Loader and Launcher creation

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: Done

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/loader/VariablesOverrider.kt
KNtronov: Done

alchemist-cognitive-agents/src/test/resources/arrive.yml
```diff
@@ -1,5 +1,8 @@
 incarnation: protelis
 
+launcher:
+  type: HeadlessSimulationLauncher
```
DanySK, line 4: What is the default launcher? I think it should be the Headless...

alchemist-web-renderer/build.gradle.kts
```diff
@@ -95,6 +95,9 @@ application {
  * Webpack task that generates the JS artifacts.
  */
 val webpackTask = tasks.named("jsBrowserProductionWebpack")
+dependencies {
+    implementation(project(mapOf("path" to ":")))
+}
```
DanySK, line 100: What is this for?

DanySK (changes requested):
We should have a default launcher set to headless. The change may also reduce *a lot* the code needed for this refactoring

alchemist-web-renderer/build.gradle.kts
KNtronov: I don't recall adding it, it could be something IDE inserted. I see no use for it so I removed it.

DanySK (changes requested):
Can you remove all the additional

```yaml
launcher:
  type: HeadlessSimulationLauncher
```

that appear in many tests, since it is now the default?

KNtronov:
Removed useless type: Headless in test config files

alchemist-grid/src/test/resources/config/00-dependencies.yml
```diff
@@ -1,4 +1,11 @@
...
+
+launcher:
+  type: DistributedExecution
+  parameters:
+    variables: []
+    distributedConfigPath: config/dependencies_test.txt
+
 remote-dependencies:
   - config/dependencies_test.txt
   - config/dependencies_test1.txt
```
DanySK, line 11:
```suggestion
  - config/dependencies_test1.txt

```

DanySK (changes requested):
Great pull request, code wise. There is only one extremely minor suggestion, plus a missing point: there is no documentation of the CLI.
Can you write the logics and the reference guide here?
`site/content/reference/cli/_index.md`

DanySK (approved):

PR #2563, opened 2023-08-24
feat(engine): parallel engine

KNtronov:
- Implement Batch Engine
- Implement Engine Configuration


alchemist-benchmark/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+    jmhImplementation("org.openjdk.jmh:jmh-core:1.36")
+    jmhImplementation("org.openjdk.jmh:jmh-generator-annprocess:1.36")
+    jmhImplementation(project(mapOf("path" to ":")))
+    jmh("org.apache.commons:commons-math3:3.6.1")
+    runtimeOnly(rootProject)
+}
+
+tasks.withType<ShadowJar>().configureEach {
+    isZip64 = true
+}
```
DanySK, line 36: I think that alchemist-benchmark should actually be on a separate repository, importing alchemist. The distribution is already very big.

DanySK (changes requested):
First batch of changes requested

alchemist-benchmark/build.gradle.kts
KNtronov: A separate git repository?

alchemist-benchmark/.gitignore
```diff
@@ -0,0 +1,42 @@
...
+/nbbuild/
+/dist/
+/nbdist/
+/.nb-gradle/
+
+### VS Code ###
+.vscode/
+
+### Mac OS ###
+.DS_Store
```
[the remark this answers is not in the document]
KNtronov: Done

alchemist-api/src/main/kotlin/it/unibo/alchemist/model/OutputReplayStrategy.kt
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (C) 2010-2023, Danilo Pianini and contributors
+ * listed, for each module, in the respective subproject's build.gradle.kts file.
+ *
+ * This file is part of Alchemist, and is distributed under the terms of the
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.model
```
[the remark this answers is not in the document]
KNtronov: Done

alchemist-benchmark/build.gradle.kts
KNtronov: Removed module, 

This is the repo https://github.com/KNtronov/alchemist-benchmarks

DanySK (changes requested):
`EngineConfiguration`, `BatchEngineConfiguration`, and `BatchedScheduler` should be in alchemist-engine.

Also, `EngineConfiguration` and `BatchEngineConfiguration` should be into the `core` package

DanySK:
@Mergifyio update

KNtronov:
> `EngineConfiguration`, `BatchEngineConfiguration`, and `BatchedScheduler` should be in alchemist-engine.
> 
> Also, `EngineConfiguration` and `BatchEngineConfiguration` should be into the `core` package

Refactored

alchemist-engine/src/main/java/it/unibo/alchemist/core/ArrayIndexedPriorityQueue.java
```diff
@@ -23,20 +23,18 @@
  *
  * @param <T> concentration type
  */
-public final class ArrayIndexedPriorityQueue<T> implements Scheduler<T> {
+public class ArrayIndexedPriorityQueue<T> implements Scheduler<T> {
```
DanySK, line 26: Can you retain final and implement by delegation?
Kotlin side, using `by`

alchemist-engine/src/main/java/it/unibo/alchemist/core/ArrayIndexedPriorityFixedBatchQueue.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Batched extension for ArrayIndexedPriorityQueue.
+ * This implementation presents fixed size batches.
+ *
+ * @param <T> concentration type
+ */
+public final class ArrayIndexedPriorityFixedBatchQueue<T> extends ArrayIndexedPriorityQueue<T> implements BatchedScheduler<T> {
```
DanySK, line 23: I would rewrite this in kotlin as

```suggestion
class ArrayIndexedPriorityFixedBatchQueue<T> : Scheduler<T> by ArrayIndexedPriorityQueue<T>(), BatchedScheduler<T> {
```

alchemist-engine/src/main/java/it/unibo/alchemist/core/BatchEngine.kt
```diff
@@ -0,0 +1,203 @@
...
+import java.util.stream.Collectors
+
+/**
+ * This class implements a simulation. It offers a wide number of static
+ * factories to ease the creation process.
+ *
+ * @param <T> concentration type
+ * @param <P> [Position] type
+</P></T> */
+class BatchEngine<T, P : Position<out P>> :
```
DanySK, line 32: Same as above: can we refactor to favor composition over inheritance?

alchemist-engine/src/main/java/it/unibo/alchemist/core/EngineConfiguration.kt
```diff
@@ -0,0 +1,15 @@
...
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.core
+
+/**
+ * Contains the engine configuration parameters.
+ */
+interface EngineConfiguration
```
DanySK, line 15: Remind me why we need this instead of constructor params

DanySK (changes requested):
Investigate the possibility of favoring composition over inheritance (see also Effective Java)

alchemist-engine/src/main/java/it/unibo/alchemist/core/EngineConfiguration.kt
KNtronov: As per Arbitrary class loading system documentation:

`Alchemist is able to load arbitrary types conforming to the expected interface (or Scala trait). The expected type depends on where the class is requested.`

Configuration is loaded with
SimulationModel.visitBuilding<EngineConfiguration>(context, engineConfigurationDescriptor)

Which resolves to the specified type ex. EpsilonBatchEngineConfiguration

alchemist-engine/src/main/java/it/unibo/alchemist/core/BatchEngine.kt
KNtronov: I'm not sure if delegation is applicable in a clean way. The way this class is designed, we override an internal piece of logic doStep().

The only option I see is to extract doStep in a function (Engine<T, P> -> Unit) and bring it into the constructor, however if we use delegation we wouldn't have a way to enforce the implementation of this function, that would leak the logic to the classes constructing an Engine.

Maybe there is some other Kotlin mechanism that would help with this issue I'm missing here?

alchemist-engine/src/main/java/it/unibo/alchemist/core/ArrayIndexedPriorityQueue.java
KNtronov: Done

alchemist-engine/src/main/java/it/unibo/alchemist/core/ArrayIndexedPriorityFixedBatchQueue.java
KNtronov: Done

alchemist-engine/src/main/java/it/unibo/alchemist/core/EngineConfiguration.kt
DanySK: But why cannot we instead have plain old constructor parameters and do:
`SimulationModel.visitBuilding<Simulation>(context, engineConfigurationDescriptor)`

It looks like `EngineConfiguration` is a container object that exists only to provide a single parameter to `Engine`

alchemist-engine/src/main/java/it/unibo/alchemist/core/BatchEngine.kt
DanySK: Uhmm... I can see the problem. You would need the delegate to return control, but to do so you need a virtual method, and that method is not declared in the interface.

Which method is being overridden?

KNtronov: protected void doStep()

It's been extracted from the main algorithm with the intention to be overriden to customize a single simulation step logic.

alchemist-engine/src/main/java/it/unibo/alchemist/core/EngineConfiguration.kt
KNtronov: I can see some issues with building the Simulation (and specifically its Engine implementation)

1. Simulation is a generics interface (Simulation<T, P extends Position<? extends P>>), how can we infer the generic types from yaml?

2. Engine has a non-simple type constructor signature:
`public Engine(final Environment<T, P> e, final long maxSteps, final Time t, final Scheduler<T> scheduler)`

Scheduler addition is new (with the intent of using composition to build a custom engine), so it can be ignored as we can construct a correct scheduler for the correct Engine type, but the Environment<T, P> is from the original implementation. Environment is not a simple properties container, it typically contains mutable objects instanced in some business logic.

> It looks like EngineConfiguration is a container object that exists only to provide a single parameter to Engine

There is a type hierarchy that progressively adds mandatory parameters:

`
interface BatchEngineConfiguration : EngineConfiguration {

    val outputReplayStrategy: String
}

class FixedBatchEngineConfiguration(
    override val outputReplayStrategy: String,
    val batchSize: Int,
) : BatchEngineConfiguration

class EpsilonBatchEngineConfiguration(
    override val outputReplayStrategy: String,
    val epsilonValue: Double,
) : BatchEngineConfiguration
`

So actually it supplies multiple parameters to the engine. Hence why I decided to map EngineConfiguration to an Engine impelentation instead of constructing an Engine from yaml.

DanySK (approved):

DanySK:
@KNtronov for some reason building the website fails -- can you check why?

KNtronov:
Looks like an issue with git module:
`Error: failed to load modules: module "hugo-theme-relearn" not found in "/home/runner/work/Alchemist/Alchemist/site/themes/hugo-theme-relearn"; either add it as a Hugo Module or store it in "/home/runner/work/Alchemist/Alchemist/site/themes".: module does not exist`

It seems like it's been de-registered despite being in the .gitmodules

KNtronov:
I re-added the submodule, pushing my commit

.github/workflows/build-and-deploy.yml
```diff
@@ -74,7 +74,7 @@ jobs:
   # Builds the website and verifies link sanity
   build-website:
     runs-on: ubuntu-22.04
-    timeout-minutes: 30
+    timeout-minutes: 60
```
DanySK, line 77:
```suggestion
    timeout-minutes: 30
```

PR #2706, opened 2023-10-08
feat(graphql): make `Environment` accessible through GraphQL query

S-furi:
This PR focuses in providing a first working implementation of a GraphQL server that mainly exposes `Environment` through its GraphQL surrogate.
Also `Layer` is now mapped onto its surrogate class.
Every element exposed through the environment (e.g. nodes, molecules, concentrations, …), is now being documented in the generated GraphQLSchema, thanks to `@GraphQLDescription`.

The GraphQL server can be now launched as every `SimulationLauncher` through the command line, and when up and running, the environment query can be tested in the playground at `http://$host:$port/graphiql`.


alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/EnvironmentSurrogateTest.kt
```diff
@@ -9,23 +9,24 @@
...
 import it.unibo.alchemist.model.Time
 import it.unibo.alchemist.model.geometry.Vector
 
 class EnvironmentSurrogateTest<T, P> : StringSpec({
     "EnvironmentSurrogate should map an Environment to a GraphQL compliant object" {
         graphqlTestEnvironmnets<T, P>().forEach {
-            val envSurrogate = it.toGraphQLEnvironmentSurrogate()
+            if (it.simulation == null) {
+                fail("Simulation is null")
+            }
```
DanySK, line 27: `requireNotNull` { ... }

alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/TestingEnvironments.kt
```diff
@@ -15,14 +15,15 @@ import it.unibo.alchemist.model.EuclideanEnvironment
...
+import it.unibo.alchemist.test.startSimulation
 import java.io.File
 
 object TestingEnvironments {
     fun <T, P> graphqlTestEnvironmnets(): Set<EuclideanEnvironment<T, P>> where P : Position<P>, P : Vector<P> =
         this::class.java.classLoader.getResource("yaml")?.path?.let { path ->
             File(path).listFiles()?.map {
                 loadYamlSimulation<T, P>("yaml/${it.name}")
-            }?.toSet()
+            }?.toSet()?.onEach { it.startSimulation() }
```
DanySK, line 26: maybe using a sequence and an executor? Otherwise it preloads everything in memory...

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/surrogates/TimeSurrogate.kt
```diff
@@ -9,14 +9,20 @@
 
 package it.unibo.alchemist.boundary.graphql.schema.model.surrogates
 
+import com.expediagroup.graphql.generator.annotations.GraphQLDescription
 import it.unibo.alchemist.model.Time
 
 /**
  * A GraphQL surrogate for a [Time].
  */
 enum class TimeSurrogate {
```
DanySK, line 18: How can this work? The time is usually a number

DanySK (changes requested):
Reviewed 21/23. @AngeloFilaseta I want your review, too

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/launchers/GraphQLServerLauncher.kt
```diff
@@ -20,21 +25,37 @@ import kotlinx.coroutines.runBlocking
...
+            Netty,
+            port = port,
+            host = host,
+            module = {
+                attributes.put(SimulationAttributeKey, simulation)
+                graphQLModule()
+                graphQLRoutingModule()
+            },
+        )
 }
```
AngeloFilaseta, line 61: Can you remember me why you're using an `embeddedServer` instead on of the ktor Engine? As far as I remember using the other method is considered the "best practice", but maybe I'm just forgetting something.

AngeloFilaseta: Is it to pass the Simulation? Can you please check if it is possible to do so using the other method too?

alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/TestingEnvironments.kt
S-furi: I've changed the code in order to use a sequence. I've tried to use executors through java `ExecutorService` but then the `addNode` method in [environment test](https://github.com/AlchemistSimulator/Alchemist/pull/2706/files#diff-f5fb2b229c40eb15a21c4707cbe855f9cbf1278967c527dd1e3569eb4d4c7d3eR40) fails, because the method should be called inside the simulation thread, and if I try to use the `Simulation.schedule` method, an exception is thrown due to the termination of the simulation. Maybe i am messing up with threads and the simulation launching procedure.

alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/EnvironmentSurrogateTest.kt
S-furi: FIxed!

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/launchers/GraphQLServerLauncher.kt
AngeloFilaseta: Ok, this can be considered resolved for me.

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/surrogates/TimeSurrogate.kt
S-furi: I have misinterpreted the `Time` class. I have already added the correct implementation.

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/launchers/GraphQLServerLauncher.kt
S-furi: @DanySK can i mark this as resolved?

alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/TestingEnvironments.kt
```diff
@@ -15,19 +15,22 @@ import it.unibo.alchemist.model.EuclideanEnvironment
...
         this::class.java.classLoader.getResource("yaml")?.path?.let { path ->
-            File(path).listFiles()?.map {
-                loadYamlSimulation<T, P>("yaml/${it.name}")
-            }?.toSet()
+            File(path).listFiles()?.asSequence()?.distinct()?.map {
+                val env = loadYamlSimulation<T, P>("yaml/${it.name}")
+                env.startSimulation()
+                env
+            }
         }.orEmpty()
```
DanySK, line 29: I would fail on an empty sequence, if there is at least one test

alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/TestingEnvironments.kt
```diff
@@ -15,19 +15,22 @@ import it.unibo.alchemist.model.EuclideanEnvironment
...
-    fun <T, P> graphqlTestEnvironmnets(): Set<EuclideanEnvironment<T, P>> where P : Position<P>, P : Vector<P> =
+    fun <T, P> graphqlTestEnvironments(): Sequence<EuclideanEnvironment<T, P>> where P : Position<P>, P : Vector<P> =
         this::class.java.classLoader.getResource("yaml")?.path?.let { path ->
-            File(path).listFiles()?.map {
-                loadYamlSimulation<T, P>("yaml/${it.name}")
-            }?.toSet()
+            File(path).listFiles()?.asSequence()?.distinct()?.map {
+                val env = loadYamlSimulation<T, P>("yaml/${it.name}")
+                env.startSimulation()
+                env
```
DanySK, line 27:
```suggestion
                loadYamlSimulation<T, P>("yaml/${it.name}").apply { startSimulation() }
```

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/launchers/GraphQLServerLauncher.kt
AngeloFilaseta: I don't know why but I could'nt. Maybe my UI was bugged, or (even more) probably i'm just blind.

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/surrogates/TimeSurrogate.kt
```diff
@@ -9,22 +9,62 @@
...
+     */
+    @GraphQLDescription("The time represented as a double")
+    val doubleTime: Double
+        get() = origin.toDouble()
+}
+
+/**
+ * A GraphQL input types used by clients for representing time.
+ */
+enum class TimeInput {
```
DanySK, line 34: What is this enum for?

DanySK (changes requested):

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/surrogates/TimeSurrogate.kt
S-furi: As long as I understood, we cannot create arbitrary time, but only through the companion object provided by `Time`. This enum is the 1-1 representation of that companion object, because GraphQL Kotlin does not consider companion objects when creating the schema.
In this way clients that has to specify a Time for some operations (e.g. `EnvironmentSurrogate.addNode`), would use this enum.

alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/TestingEnvironments.kt
S-furi: If the sequence is empty (i.e. no yaml files were found in the `resource/yaml` folder),  the [`graphqlTestEnvironments` test](https://github.com/AlchemistSimulator/Alchemist/blob/9d4287d8ff8e4ec6ff083193dc2c3f40ab8a0cb0/alchemist-graphql/src/jvmTest/kotlin/it/unibo/alchemist/boundary/TestingEnvironments.kt#L34) will fail. I do not know if I understood correctly your point.

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/surrogates/TimeSurrogate.kt
S-furi: @DanySK does it make sense?

DanySK:
@S-furi let me know when this one is ready for another spin

S-furi:
@DanySK if you think [this](https://github.com/AlchemistSimulator/Alchemist/pull/2706#discussion_r1352682234) is ok, it is ready!

DanySK:
I still do not understand that enum

DanySK:
Can you simply say that time is backed by a double value?

DanySK:
I would remove the enum for now, if it does not break any query

S-furi:
The only problem now is when [cloning a node](https://github.com/AlchemistSimulator/Alchemist/blob/e3341779ed8948f673d0b0caedca26c0ac2afa87/alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/model/surrogates/EnvironmentSurrogate.kt#L98) i have no clue on how to properly map the given `TimeSurrogate` to its Alchemist `Time` representation, because `Time` or `DoubleTime` do not provide ways to create instances of time with an arbitrary Double value, but they let the creation through the companion object, and this is the motivation behind the enum (it emulates the behaviour of the companion object).
So yes, removing the enum does break the node cloning operation.

DanySK:
Why not `suspend fun cloneNode(nodeId: Int, position: PositionInput, time: Double): NodeSurrogate<T>?)`?
I struggle to understand where the issue is, maybe we take a look at it together?

S-furi:
This works if only when we clone the original Alchemist Node we do something like `val newNode = origin.getNodeById(nodeId).cloneNode(DoubleTime(time))`. If this is ok I will proceed in changing the code and removing the enum!

DanySK:
> This works if only when we clone the original Alchemist Node we do something like `val newNode = origin.getNodeById(nodeId).cloneNode(DoubleTime(time))`. If this is ok I will proceed in changing the code and removing the enum!

looks fit to me: what was the issue? Invoking `DoubleTime` directly?

S-furi:
FIxed it! I was getting confused by time creation and thought the only way was through the companion object, and also I missed the secondary constructor of `DoubleTime(time: Double)`.
Now everything should be fine and nothing breaks!

DanySK (approved):

PR #2832, opened 2023-11-07
feat(graphql): add common GraphQL client

S-furi:
This  PR contains a common multiplatform GraphQL client implementation, which uses Apollo GraphQL under the hood. A factory of clients is made available, making possibile to choose between a standard GraphQL client (more lightweight which does not provides support for subscriptions) and a client that fully supports subscriptions through WebSocket.
The provided code also include the final GraphQL code generated by surrogates used by Apollo's clients.
Finally, a simple simulation handler has been created, which can perform Play/Pause/Terminate client side mutations upon loaded simulation on the server.


alchemist-graphql/build.gradle.kts
```diff
@@ -93,12 +87,7 @@ kotlin {
                 implementation(alchemist("test"))
             }
         }
-        val jsMain by getting {
-            dependencies {
-                implementation(libs.apollo.runtime)
-                implementation(libs.kotlin.coroutines.core)
-            }
-        }
+        val jsMain by getting
```
DanySK, line 90: it looks like this sourceset is declared but contains  no source

alchemist-graphql/build.gradle.kts
```diff
@@ -124,6 +141,15 @@ tasks.named("run", JavaExec::class).configure {
     )
 }
 
+/**
+ * Configure GraphQL Schema download task to download the schema from the server
+ * at the default endpoint, locating the updated schema under the correct directory.
+ */
+tasks.withType<com.expediagroup.graphql.plugin.gradle.tasks.GraphQLDownloadSDLTask>().configureEach {
+    outputFile.set(File("src/commonMain/resources/graphql/schema.graphqls"))
+    endpoint = "http://localhost:8081/sdl"
```
DanySK, line 150: what is sdl for?

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/DefaultGraphQLClient.kt
```diff
@@ -0,0 +1,98 @@
...
+            return DefaultGraphQLClient(this)
+        }
+    }
+}
+
+/**
+ * Exception thrown when the client is not built properly.
+ */
+class ClientNotBuiltException :
+    Exception("The client was not built properly! Make sure to call `build()` before running any operation.")
```
DanySK, line 98: No custom exceptions

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/DefaultGraphQLClient.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    override fun <D : Subscription.Data> subscription(subscription: Subscription<D>): ApolloCall<D> {
+        checkBuilt()
+        check(isSubscriptionModuleEnabled) { "Subscription module is not enabled!" }
+        return client.subscription(subscription)
+    }
+
+    override fun close() = client.close()
+
+    private fun checkBuilt() = if (!isBuilt) throw ClientNotBuiltException() else true
```
DanySK, line 53: just check(isBuilt) { "..." }

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/DefaultGraphQLClient.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.apollographql.apollo3.api.Query
+import com.apollographql.apollo3.api.Subscription
+import com.apollographql.apollo3.network.ws.GraphQLWsProtocol
+import com.apollographql.apollo3.network.ws.WebSocketNetworkTransport
+
+/**
+ * Default GraphQL client implementation.
+ * @param client the Apollo client
+ */
+class DefaultGraphQLClient(private val client: ApolloClient) : GraphQLClient {
```
DanySK, line 24: This is a strange piece of engineering: the outer class knows its builder. Can you elaborate on the design choice?

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/GraphQLClient.kt
```diff
@@ -0,0 +1,89 @@
...
+     * Closes the client.
+     */
+    fun close(): Unit
+}
+
+/**
+ *
+ * Builds the Apollo GraphQL client with the specified features.
+ */
+interface GraphQLClientBuilder {
```
DanySK, line 60: Split in another file please

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/GraphQLClientFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+     * @param host the host of the GraphQL server
+     * @param port the port of the GraphQL server
+     */
+    fun basicClient(
+        host: String = "127.0.0.1",
+        port: Int = 8081,
+    ): GraphQLClient =
+        DefaultGraphQLClient.Builder()
+            .serverUrl(host, port)
+            .build()
```
DanySK, line 29:
```suggestion
    ): GraphQLClient = DefaultGraphQLClient.Builder().serverUrl(host, port).build()
```

alchemist-graphql/src/jvmMain/kotlin/it/unibo/alchemist/boundary/graphql/schema/operations/subscriptions/util/SubscriptionsUtils.kt
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Returns the [EnvironmentSubscriptionMonitor] of the given [Environment].
+     * @param environment the environment.
+     */
+    fun <T, P>environmentSubscriptionMonitor(environment: Environment<T, P>): EnvironmentSubscriptionMonitor<T, P>
+        where P : Position<out P> =
+        environment.simulation.outputMonitors.filterIsInstance<EnvironmentSubscriptionMonitor<T, P>>()
+            .apply { check(size == 1) { "Only one subscription monitor is allowed" } }
+            .first()
+}
```
DanySK, line 30: No "utils". If there are util-like functionalities needed, it is better to prepare extension packs (see how it's done in the remainder of the project)

```suggestion
object Environments {

    /**
     * Returns the [EnvironmentSubscriptionMonitor] of the given [Environment].
     * @param environment the environment.
     */
    fun <T,  P : Position<out P>> Environment<T, P> subscriptionMonitor() : EnvironmentSubscriptionMonitor<T, P> =
        simulation.outputMonitors.filterIsInstance<EnvironmentSubscriptionMonitor<T, P>>()
            .apply { check(size == 1) { "Only one subscription monitor is allowed" } }
            .first()
}
```

DanySK (changes requested):
Gave a look, let's implement the first round of changes and then re-review

alchemist-graphql/build.gradle.kts
S-furi: it stands for Schema Definition Language, and `GraphQLDownloadSDL` task downloads the schema at the provided endpoint.

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/DefaultGraphQLClient.kt
S-furi: The reason behind this design was to enforce the use use of builder, much like Apollo Client API make their clients being built. If this is bad, I will change it in a standard builder.

DanySK: A classic (non step-) Builder serves two main purposes:
1. support for many parameters with a default
2. provide a clear indication of what some parameter does

It is thus particularly useful when the target language has no default and named parameter (like Java) and a bit less useful when the language does (Kotlin, Scala). Is Apollo a Java library? Or is it Kotlin-native?

S-furi: Apollo is written mainly Kotlin, but I got your point and so the Builder isn't really necessary; I can remove the builder and build it through the factory directly.

S-furi: I removed the builder and use only named parameters and the factory.

alchemist-graphql/build.gradle.kts
```diff
@@ -66,6 +57,9 @@ kotlin {
         val commonMain by getting {
             dependencies {
                 implementation(libs.kotlinx.serialization.json)
+                implementation(libs.kotlin.stdlib)
+                implementation(libs.apollo.runtime)
+                implementation(libs.kotlin.coroutines.core)
```
DanySK, line 62: sort lexicographically

alchemist-graphql/build.gradle.kts
```diff
@@ -93,20 +87,41 @@ kotlin {
...
 application {
     mainClass.set("it.unibo.alchemist.Alchemist")
 }
 
+/**
+ * Configure the Apollo Gradle plugin to generate Kotlin models
+ * from the GraphQL schema inside the `commonMain` sourceSet.
+ */
+apollo {
+    service("alchemist") {
```
DanySK, line 102:
```suggestion
    service(name) {
```

alchemist-graphql/build.gradle.kts
```diff
@@ -93,20 +87,41 @@ kotlin {
...
+        generateKotlinModels.set(true)
+        packageName.set("it.unibo.alchemist.boundary.graphql.client")
+        schemaFiles.from(file("src/commonMain/resources/graphql/schema.graphqls"))
+        srcDir("src/commonMain/resources/graphql")
+        outputDirConnection {
+            connectToKotlinSourceSet("commonMain")
```
DanySK, line 108: @AngeloFilaseta wouldn't it be better to declare a new `generatedCommonMain` sourceset? What do you think?

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/SimulationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Plays the simulation.
+     */
+    suspend fun play(): PlaySimulationMutation.Data {
+        val status = handleSimulation(
+            PlaySimulationMutation(),
+            "STARTING SIMULATION",
+        ) as PlaySimulationMutation.Data
+
+        require(status.play == "RUNNING") { status.play }
```
DanySK, line 48:
```suggestion
        check(status.play == "RUNNING") { status.play }
```

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/SimulationHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+        return status
+    }
+
+    /**
+     * Returns the status of the simulation.
+     */
+    suspend fun status() = graphqlClient.query(SimulationStatusQuery()).execute().data?.simulationStatus
+
+    private suspend fun handleSimulation(action: Mutation<*>, message: String) =
+        graphqlClient.mutation(action).execute().also { println(message) }.data!!
```
DanySK, line 58: Use the logger, never `println` stuff

alchemist-graphql/src/commonMain/resources/graphql/schema.graphqls
```diff
@@ -0,0 +1,185 @@
...
+"A molecule with an associated name, used as input object"
+input MoleculeInput {
+  name: String!
+}
+
+"Position in space, used as input parameter"
+input PositionInput {
+  coordinates: [Float!]!
+  dimensions: Int!
+}
```
DanySK, line 185: This looks generated... Is it?

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/DefaultGraphQLClient.kt
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Default GraphQL client implementation.
+ *
+ * @param host the host of the GraphQL server
+ * @param port the port of the GraphQL server
+ * @param enableSubscription whether to enable subscriptions or not
+ */
+class DefaultGraphQLClient(
+    private val host: String = "127.0.0.1",
+    private val port: Int = 8081,
```
DanySK, line 29: This number occurs too many times. maybe create a constant in a (companion?) object

DanySK (changes requested):
Reviewed 7/7. A few changes requested

alchemist-graphql/src/commonMain/resources/graphql/schema.graphqls
S-furi: This is downloaded from the server endpoint while it is up and running and it is needed at compile time in for the client.

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/SimulationHandler.kt
S-furi: Which logger could I use considering that I'm in common code?

alchemist-graphql/src/commonMain/resources/graphql/schema.graphqls
DanySK: So it's generated, is it correct? If so, we could write a gradle task that generates these files (it should be there already, actually), and then alter the classpath of the client to include these files

alchemist-graphql/src/commonMain/kotlin/it/unibo/alchemist/boundary/graphql/client/SimulationHandler.kt
DanySK: That's an issue indeed... do we *need* that logger call?

S-furi: Actually we do not need it here! I'll remove it by now, and when a multiplatform logger will be available I can adapt the code accordingly.

alchemist-graphql/src/commonMain/resources/graphql/schema.graphqls
S-furi: Yes it is generated, but it can be downloaded only when the server is up, and the client requires the schema to be present when it is being compiled.

There is a task called `graphqlGenerateSDL` that generate the schema at compile time, but it makes use of `NoopSchemaGeneratorHooks` which, in a few words, cannot generate the schema in our case because it is not designed to support Subscriptions that use Kotlin `Flow`s. 
For those operation we need `FlowSubscriptionSchemaGeneratorHooks` which is used inside the server, and so the schema is properly generated at runtime, but at compile time that set of hooks are not available inside the GraphQL Kotlin gradle plugin.

I have searched in the [documentation](https://opensource.expediagroup.com/graphql-kotlin/docs/plugins/gradle-plugin-tasks#graphqlgeneratesdl) how to add a set of custom hooks that can be used by the gradle plugin, but it is quite complex to do (see [the guide](https://opensource.expediagroup.com/graphql-kotlin/docs/plugins/hooks-provider#creating-custom-hooks-service-provider)).
In case, I could raise an issue at GraphQL Kotlin, pointing out that they could provide `FlowSubscriptionSchemaGeneratorHooks` support inside the gradle plugin, as they do with federated schemas.
Otherwise I can try to build a custom one that exposes the flow subscription hooks.

DanySK: Can we start the server and then grab those files before the compilation of the client?
The server is in a separate project, isn't it? We could have a task that launches it for a snap, making it generate the files, and then we instruct the compiler to add them to the compile classpath.

Not a trivial Gradle exercise, but I believe it's doable. Maybe we can try to do it together

alchemist-graphql/build.gradle.kts
AngeloFilaseta: If this is inevitable, separation can certainly help. I would also add some indication somewhere (a readme in the sourceset or something similar) to ensure that developers can tedit what's inside only if they are really sure of what they're doing.

DanySK:
@S-furi can you write a guide to graphql in site?
@AngeloFilaseta can this one get merged? We should have the generation implemented, don't we?

S-furi:
Yes, I was wondering on how to do it. Are there guidelines on how to add documentation in the website?

AngeloFilaseta:
@DanySK yes , I think we should be ready to merge this 😄 
@S-furi I'm aware that Alchemist uses the [DIVIO Documentation System](https://documentation.divio.com/).

S-furi:
Got it!

alchemist-graphql/src/commonMain/resources/graphql/schema.graphqls
DanySK: Should be done, right @AngeloFilaseta and @S-furi ?

alchemist-graphql/build.gradle.kts
DanySK: This should be fixed also, @AngeloFilaseta confirm and resolve in case

alchemist-graphql/src/commonMain/resources/graphql/schema.graphqls
S-furi: Yes to me this is fixed!

DanySK:
So we are only waiting for the docs to merge, correct?

DanySK (approved):

DanySK:
@S-furi , can you create a new PR with the documentation? I am merging this one in the meanwhile.

PR #3022, opened 2024-01-04
feat: release platform-specific packages

Zimbrando:
This PR adds the automatic packaging in the CICD pipeline. Packages are tested by the "test-packaging" job consisting in the creation and file integrity checks. JPackage is used for this purpose.


.github/workflows/build-and-deploy.yml
```diff
@@ -88,6 +88,41 @@ jobs:
...
+        uses: actions/setup-java@v4.0.0
+        with:
+          java-version: ${{ needs.select-java-version.outputs.java-version }}
+          distribution: 'temurin'
+      - name: Dry run
+        shell: bash
+        run: ./gradlew jpackage --info -Djpackage.dryRun=true
+      - name: Install bsdtar
+        if: matrix.os == 'ubuntu-22.04'
+        run: sudo apt update && sudo apt install libarchive-tools
```
DanySK, line 121: Is there a container that can do this instead?

.github/workflows/build-and-deploy.yml
```diff
@@ -88,6 +88,41 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ windows-2022, macos-12, ubuntu-22.04 ]
+    runs-on: ${{ matrix.os }}
+    if: >-
+      !github.event.repository.fork
+      && (
+        github.event_name != 'pull_request'
+        || github.event.pull_request.head.repo.full_name == github.repository
+      )
```
DanySK, line 104: why do we need to exclude the forks if no secret is required to package?

.github/workflows/build-and-deploy.yml
```diff
@@ -165,6 +200,7 @@ jobs:
       - build
       - build-website
       - test-deploy
+      - test-packaging
```
DanySK, line 203: Why do we need to package to build the website?

package-settings/windows/main.wxs
```diff
@@ -0,0 +1,127 @@
...
+      <Custom Action="JpDisallowUpgrade" After="FindRelatedProducts">JP_UPGRADABLE_FOUND</Custom>
+      <?endif?>
+      <?ifndef JpAllowDowngrades ?>
+      <Custom Action="JpDisallowDowngrade" After="FindRelatedProducts">JP_DOWNGRADABLE_FOUND</Custom>
+      <?endif?>
+      <RemoveExistingProducts Before="CostInitialize"/>
+    </InstallExecuteSequence>
+
+  </Product>
+</Wix>
```
DanySK, line 127: This file looks generated. What is it?

build.gradle.kts
```diff
@@ -39,6 +43,7 @@ plugins {
     alias(libs.plugins.shadowJar)
     alias(libs.plugins.taskTree)
     alias(libs.plugins.hugo)
+    alias(libs.plugins.jpackage)
```
DanySK, line 46: Shouldn't we apply this solely to `alchemist-full`?

DanySK (changes requested):
First, I am pleased with the quality of this work.
I've got several questions though, so please try to interact in the conversations and notify me as soon as all of them are resolved, so that I can run another review round

package-settings/windows/main.wxs
Zimbrando: Initially I was packaging all the JARs producing a .msi installer of over 2GB. Because of a limit by design for .cab files it was not working. Now we are packaging only the full-jar and it's way lighter, probably we don't need this anymore. I'll test it again and remove it if useless. If ever in the future the package gets heavier this fix it's the only one that i've found. It is generated by the --temp attribute in jpackage that saves all the temporary files in a directory specified by the user. Then, i've modified the <Media> element to a <MediaTemplate> one.

Further info here: 
[cpack-msi-installation-exceeding-2gb-packing-into-multiple-cab-files](https://stackoverflow.com/questions/62279584/cpack-msi-installation-exceeding-2gb-packing-into-multiple-cab-files)
[override-jpackage-resources](https://docs.oracle.com/en/java/javase/21/jpackage/override-jpackage-resources.html#GUID-1B718F8B-B68D-4D46-B1DB-003D7729AAB6)

build.gradle.kts
Zimbrando: Applying and configuring all the jpackage related stuff in the alchemist-full/build.gradle.kts?

.github/workflows/build-and-deploy.yml
Zimbrando: The 'test-packaging' job is necessary for the 'release' job, not for the 'build-website' job

.github/workflows/build-and-deploy.yml
Zimbrando: I've copied it from "test-deploy" without thinking. At first I thought test-deploy was disabled for other reasons, the only downside that comes to my mind is the higher execution time of the pipeline with test-packaging enabled. I'll remove the fork check

build.gradle.kts
DanySK: yes

.github/workflows/build-and-deploy.yml
DanySK: then specify it as a dependency for release, not for build-website

.github/workflows/build-and-deploy.yml
DanySK: resolve when done

package-settings/windows/main.wxs
DanySK: if it's generated, it should never get tracked. Please check whether we need it at all

DanySK:
@Zimbrando is it okay with you if I rebase this branch automatically? If you do not know how to use rebase (which is legit) then I will automatically update with usual merges. Let me know.

Zimbrando:
> @Zimbrando is it okay with you if I rebase this branch automatically? If you do not know how to use rebase (which is legit) then I will automatically update with usual merges. Let me know.

Ok. Yes, I have never used git rebase before.

.github/workflows/build-and-deploy.yml
Zimbrando: It's already specified in the "needs" section of release and not build-website. 
Github with this view made a mess, + "- test-packaging" is in line 203 while build-website needs are in lines 127-128

https://github.com/Zimbrando/Alchemist/blob/911411336d985e239d5280273225195808f76ca3/.github/workflows/build-and-deploy.yml#L203

DanySK:
Then zero risk, let us use merge. I will squash the changes in mainline later.

build.gradle.kts
Zimbrando: Shall I upload in the release all the JARs with the packages or only the packages? 
Packages provides an easy way to install and use alchemist-full. If the other modules' jar are useful we could upload the packages + all the other jars except the full one.

DanySK: At the moment I'd keep the current jars + the redistributable packages. It'd make sense, though, to ship only alchemist-full, as jar or platform-specific redistributable.

Zimbrando:
> @Zimbrando can you "allow editings from the maintainers"? Otherwise we cannot auto-merge, and it is your responsibility to update the project every time a new commit hits the Alchemist's master branch.

image

It was already set to allowed, I unchecked it and checked it again

DanySK:
Then uhmm... Can you install mergify on your profile?

Zimbrando:
Done

DanySK:
okay, it looks like the auto-merge works.

DanySK:
@anitvam do you mind providing a second opinion once I'm satisfied with the PR?

.github/workflows/build-and-deploy.yml
Zimbrando: Something like [this](https://hub.docker.com/r/tim03/libarchive)? 
Unfortunately github removed libarchive-tools from the ubuntu vm and I did not find any alternative tool to extract a rpm package. I tried also rpm2cpio | cpio, but it was way slower and did not work from gradle.

build.gradle.kts
```diff
@@ -304,6 +305,7 @@ allprojects {
 
     // Shadow Jar
     tasks.withType<ShadowJar> {
+
```
DanySK, line 308:
```suggestion
```

build.gradle.kts
```diff
@@ -26,6 +26,7 @@ import org.jetbrains.dokka.gradle.AbstractDokkaParentTask
 import org.jetbrains.dokka.gradle.DokkaCollectorTask
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
+
```
DanySK, line 29:
```suggestion
```

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+            execFiles = workingDir.resolve("install").listFiles().map { it.name }
+            appFiles = workingDir.resolve("install/app").listFiles().map { it.name }
+        } else if (isMac) {
+            val root = File("/Applications/${rootProject.name}.app")
+            execFiles = root.resolve("Contents/MacOS").listFiles().map { it.name }
+            appFiles = root.resolve("Contents/app").listFiles().map { it.name }
+        } else if (isLinux) {
+            execFiles = workingDir.resolve("install/opt/alchemist/bin").listFiles().map { it.name }
+            appFiles = workingDir.resolve("install/opt/alchemist/lib/app").listFiles().map { it.name }
+        }
```
DanySK, line 120: this looks very clumsy and copy/pasted. Why does MacOS require a position in the file system root, while the others don't?

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+        require(rootProject.name in execFiles || "${rootProject.name}.exe" in execFiles)
+        require(tasks.jpackage.get().mainJar in appFiles)
+    }
+
+    dependsOn(tasks.jpackage)
+    finalizedBy("deleteJpackageOutput")
+}
+
+tasks.register<Delete>("deleteJpackageOutput") {
```
DanySK, line 129: val deleteJpackageOutput by tasks.registering<Delete> {

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+        } else if (isLinux) {
+            execFiles = workingDir.resolve("install/opt/alchemist/bin").listFiles().map { it.name }
+            appFiles = workingDir.resolve("install/opt/alchemist/lib/app").listFiles().map { it.name }
+        }
+        require(rootProject.name in execFiles || "${rootProject.name}.exe" in execFiles)
+        require(tasks.jpackage.get().mainJar in appFiles)
+    }
+
+    dependsOn(tasks.jpackage)
+    finalizedBy("deleteJpackageOutput")
```
DanySK, line 126: Move the delete task definition above and then reference it directly, not via strings

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+    doFirst {
+        // Extract the packet
+        if (isWindows) {
+            commandLine("msiexec", "-i", "${rootProject.name}-$version.msi", "-quiet", "INSTALLDIR=${workingDir.path}\\install")
+        } else if (isMac) {
+            commandLine("sudo", "installer", "-pkg", "${rootProject.name}-$version.pkg", "-target", "/")
+        } else if (isLinux) {
+            workingDir.resolve("install").mkdirs()
+            commandLine("bsdtar", "-xf", "${rootProject.name}-$version-1.x86_64.rpm", "-C", "install")
+        }
```
DanySK, line 104: use kotlin's `when`

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+    into("${rootProject.projectDir}/build/package-input")
+    dependsOn("shadowJar")
+}
+
+tasks.register<Exec>("testJpackageOutput") {
+    group = "Verification"
+    description = "Verifies the jpackage output correctness for the OS running the script"
+    val interceptOutput = ByteArrayOutputStream()
+    val interceptError = ByteArrayOutputStream()
+    val version = rootProject.version.toString().substringBefore('-')
```
DanySK, line 89: this must be moved in the doFirst

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+    val isLinux = isUnix && !isMac
```
DanySK, line 90: No need for this. It's the `else` in `when`

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+    from("${rootProject.projectDir}/build/shadow/$jarFile")
+    into("${rootProject.projectDir}/build/package-input")
+    dependsOn("shadowJar")
+}
+
+tasks.register<Exec>("testJpackageOutput") {
+    group = "Verification"
+    description = "Verifies the jpackage output correctness for the OS running the script"
+    val interceptOutput = ByteArrayOutputStream()
+    val interceptError = ByteArrayOutputStream()
```
DanySK, line 88: why do you intercept output and error?

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+    }
+    mac {
+        icon = "${rootProject.projectDir}/package-settings/logo.png"
+        type = ImageType.PKG
+    }
+
+    dependsOn("copyForPackaging")
+}
+
+tasks.register<Copy>("copyForPackaging") {
```
DanySK, line 77:
```suggestion
val copyForPackaging by tasks.registering<Copy> {
```

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+        winDirChooser = true
+        winShortcutPrompt = true
+        winPerUserInstall = isInCI
+    }
+    mac {
+        icon = "${rootProject.projectDir}/package-settings/logo.png"
+        type = ImageType.PKG
+    }
+
+    dependsOn("copyForPackaging")
```
DanySK, line 74: move the task above and refer it directly, not via strings

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+}
+
+tasks.register<Copy>("copyForPackaging") {
+    val jarFile = "alchemist-full-${rootProject.version}-all.jar"
+    from("${rootProject.projectDir}/build/shadow/$jarFile")
+    into("${rootProject.projectDir}/build/package-input")
+    dependsOn("shadowJar")
```
DanySK, line 81:
```suggestion
    dependsOn(tasks.shadowJar)
```

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
     mainClass.set("it.unibo.alchemist.Alchemist")
 }
 
+// JPackage
+tasks.jpackage {
+    // General info
+    resourceDir = "${rootProject.projectDir}/package-settings"
```
DanySK, line 43: Move the folder inside alchemist-full instead of referencing it from the root

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +37,99 @@ application {
...
+
+    // Packaging settings
+    input = rootProject.layout.buildDirectory.map { it.dir("package-input") }.get().asFile.path
+    destination = rootProject.layout.buildDirectory.map { it.dir("package") }.get().asFile.path
+    mainJar = "alchemist-full-${rootProject.version}-all.jar"
+    mainClass = "it.unibo.alchemist.Alchemist"
+
+    linux {
+        icon = "${rootProject.projectDir}/package-settings/logo.png"
+        type = ImageType.RPM
```
DanySK, line 60: Which alternatives to we have? RPM, MSI, PKG, and what else?

.github/workflows/build-and-deploy.yml
```diff
@@ -88,6 +88,41 @@ jobs:
...
+      - name: Setup Java
+        uses: actions/setup-java@v4.0.0
+        with:
+          java-version: ${{ needs.select-java-version.outputs.java-version }}
+          distribution: 'temurin'
+      - name: Dry run
+        shell: bash
+        run: ./gradlew jpackage --info -Djpackage.dryRun=true
+      - name: Install bsdtar
+        if: matrix.os == 'ubuntu-22.04'
```
DanySK, line 116: This is very fragile.

```suggestion
        if: runner.os == 'Linux'
```

DanySK (changes requested):

alchemist-full/build.gradle.kts
Zimbrando: Linux: 
 - rpm
 - deb

Windows:
- msi
- exe

MacOs:
- pkg
- dmg

alchemist-full/build.gradle.kts
Zimbrando: Dead code from old implementations, i'm going to remove it

alchemist-full/build.gradle.kts
Zimbrando: In Windows and Linux using "msiexec" and "bsdtar" we install/extract in a specific directory ("build/package/install") while in MacOS I did not find a way to tell the "installer" where to install the package. That is why the package is automatically installed in the Applications/ directory of the vm.

alchemist-full/build.gradle.kts
DanySK: Is there any reason not to generate all of them?

alchemist-full/build.gradle.kts
Zimbrando: We could set the installation directory from the jpackage configuration: 
 - If you run the testJpackageOutput task, the jpackage task will run with an arbitrary installation directory
 - If you run the jpackage task directly it will use the standard path

alchemist-full/build.gradle.kts
Zimbrando: I do not think the plugin supports multiple types. In order to achieve it we should run the task, reconfigure it by changing the type property then re-run it

DanySK: Is there a task type that we can configure? If so, we can generate our custom tasks.

alchemist-full/build.gradle.kts
DanySK: This is unclear to me. The task should produce an installable package. The installable package should be capable to install without Gradle being involved on arbitrary directories, or, in case of Linux, have pre-defined absolute paths.
Here, I see two relative and two absolute paths.

It's not something related to build, I need to understand *exactly* what these parameters are for.

alchemist-full/build.gradle.kts
```diff
@@ -30,6 +35,98 @@ application {
...
 
+val copyForPackaging by tasks.registering(Copy::class) {
+    val jarFile = "alchemist-full-${rootProject.version}-all.jar"
+    from("${rootProject.projectDir}/build/shadow/$jarFile")
+    into("${rootProject.projectDir}/build/package-input")
+    dependsOn(tasks.shadowJar)
+}
+
+// JPackage
+tasks.jpackage {
```
DanySK, line 46: get the task type of JPackage: we can create our own custom jpackage tasks with every generable output

alchemist-full/build.gradle.kts
Zimbrando: > The installation directory is platform-specific:
>  - On Linux, the default is /opt/application-name
>  - On macOS, the default is /Applications/application-name
>  - On Windows, the default is c:\Program Files\application-name; if the --win-per-user-install option is used, the default is C:\Users\user-name\AppData\Local\application-name
[source](https://docs.oracle.com/en/java/javase/21/jpackage/basic-packaging.html#GUID-EB1303BE-292F-4FEE-B91E-3C9CC367324B)


In the current configuration, every installable package is set to the default path. Windows has the property `winDirChooser = true`. Consequently, when you double-click the installer, it opens the classic installation GUI that allows you to modify the path where the application will be installed.

In the test task, I intentionally enforced the installation into a temporary "install" directory in the following ways:
- On Windows, by setting the variable INSTALLDIR.
- On Linux, by using the -C option in `bsdtar`.
- On MacOS, I could not find a way to force the installation into a custom directory. Therefore, the `installer` command automatically installs the package in `/Applications/alchemist`. 

I opted for using a temporary directory solely for convenience.
I hope this clarification makes everything clearer.

Zimbrando:
In the last commit, I added the generation of all possible packages per operating system (OS). I extended the `JPackageTask `class and overrode the action method to repeat itself for each possible package type. Since most of the methods are private, I believed it was a good idea to override the action method without touching the task implementation.

Currently, the test involves only one package type per OS:

- RPM for Linux
- MSI for Windows
- PKG for macOS

While adding tests for the new packages (DEB, EXE, DMG), I realized the code was becoming even clumsier. So, before attempting to find a possible solution, I thought it would be better to ask for input. Tasks of type Exec support the execution of only one command at a time.

I propose having one task per package type, which extracts, checks its integrity, and uninstalls the application. `TestJPackageOutput `could be the entry point for the tests, and depending on the OS that launched the task, the correct tests are executed.
@DanySK

PR #3079, opened 2024-01-26
ci: speed-up the workflow by uploading and downloading the shadowJars instead of regerating them

Zimbrando:
The PR updates jars and packages handling inside the workflow. The Build job generates both the jars and the packages and uploads them as artifacts.
The generation and uploading of JARs are exclusively performed by the Linux build job to prevent unnecessary overwrites.


Zimbrando:
If I understood correctly the build command inside the "build-check-deploy-gradle-action" generates the fatJars, but everything is cleaned out when finished by the "clean-command".

https://github.com/AlchemistSimulator/Alchemist/blob/4da01d24ddb53037b1104f298f01da5cd1ab5281/.github/workflows/build-and-deploy.yml#L39-L62

Wouldn't it be better if we ran shadowJar and jpackageFull in the "build-command" with the "clean-command" disabled. The next steps upload the artifacts and then we could launch the clean task manually with another step.

.github/workflows/build-and-deploy.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ windows-2022, macos-12, ubuntu-22.04 ]
+        os: [ windows-2022, macos-12, ubuntu-22.04]
```
DanySK, line 26:
```suggestion
        os: [ windows-2022, macos-12, ubuntu-22.04 ]
```

DanySK (approved):
just a spacing error, then we're good to go

PR #3081, opened 2024-01-27
feat(cd): automatic AUR release

Zimbrando:
This PR includes:
 - A new gradle task which outputs a valid PKGBUILD script depending on the current rootProject version
 - "test-aur-package" job that installs and asserts that alchemist is correctly installed
 - Steps to update the AUR package when a new release is out

As soon as #3079 will be merged, this PR needs to update accordingly by integrating artifacts downloads 

The package has been already manually uploaded ([link](https://aur.archlinux.org/packages/alchemist))


.github/workflows/build-and-deploy.yml
```diff
@@ -272,8 +308,19 @@ jobs:
...
+          echo "VERSION=$(./gradlew properties | grep ^version: | cut -d" " -f2)" >> $GITHUB_ENV
+      - name: Publish AUR package
+        uses: KSXGitHub/github-actions-deploy-aur@v2.7.0
+        with:
+          pkgname: alchemist
+          pkgbuild: build/package/PKGBUILD
+          commit_username: ${{ secrets.AUR_USERNAME }}
+          commit_email: ${{ secrets.AUR_EMAIL }}
+          ssh_private_key: ${{ secrets.AUR_SSH_PRIVATE_KEY }}
+          commit_message: Update to version ${{ env.VERSION }}
```
DanySK, line 323: this stuff should be executed by release.config.js

alchemist-full/build.gradle.kts
```diff
@@ -145,6 +146,43 @@ tasks.register<Exec>("testJpackageOutput") {
     finalizedBy(deleteJpackageOutput)
 }
 
+val generatePKGBUILD by tasks.registering(Exec::class) {
```
DanySK, line 149:
```suggestion
val generatePKGBUILD by tasks.registering<Exec> {
```

alchemist-full/build.gradle.kts
```diff
@@ -145,6 +146,43 @@ tasks.register<Exec>("testJpackageOutput") {
...
+    description = "Generates a valid PKGBUILD by replacing values in the template file"
+    val inputFile = file("${project.projectDir}/PKGBUILD.template")
+    val outputDir = file(rootProject.layout.buildDirectory.map { it.dir("package") }.get().asFile.path)
+    val tokenToReplace = "{%}"
+    var replacementValues: List<String>
+    val testSourceParam = System.getProperty("generatePKGBUILD.testSource", "false")
+    val interceptOutput = ByteArrayOutputStream()
+    standardOutput = interceptOutput
+    doFirst {
+        require(!isWindows && !isMac)
```
DanySK, line 160: Why? Generating a pkgbuild does not seem to be something requiring a specific OS

DanySK (changes requested):

.github/workflows/build-and-deploy.yml
Zimbrando: Both the publishing and script generation steps? 
Can I run an action step and have access to the repository secrets in release.config.js?

Zimbrando: We could set environmental variables with the secrets in the release job.
```
env:
  aur_username: ${{ secrets.AUR_USERNAME }}
  aur_commit_mail: ${{ secrets.AUR_COMMIT_MAIL }}
```
Then publishCmd can run a custom script that does what github-actions-deploy-aur@v2.7.0 does using those variables.

`./publishToAUR.sh`

Environmental  variables should be accessible from `publisCmd`, shouldn't they?

alchemist-full/build.gradle.kts
Zimbrando: The `<Exec>` version gives me an error.
After looking in the documentation I found tasks configured like this:
`val {name} by tasks.registering(Exec::class) {`
or 
`tasks.register<Exec>("{name}") {`

.github/workflows/build-and-deploy.yml
DanySK: > Both the publishing and script generation steps? Can I run an action step and have access to the repository secrets in release.config.js?

the script generation and testing can happen always. The publication only when there is a release.

You cannot run an action step in release.config.js, unfortunately. We will need a workaround.

Passing secrets is not a big deal (it's being done already). There is documentation on the build-and-deploy-gradle action on how to pass custom secrets.

.github/workflows/build-and-deploy.yml
```diff
@@ -142,6 +142,50 @@ jobs:
...
         shell: bash
         run: ./gradlew testJpackageOutput --info     
+  # Test PKGBUILD for aur release
+  test-aur-package:
+    needs: 
+      - select-java-version
+      - build
+    runs-on:  ubuntu-22.04
+    container: 
+      image: archlinux:base-devel
```
DanySK, line 152: use a fixed version

.github/workflows/build-and-deploy.yml
```diff
@@ -142,6 +142,50 @@ jobs:
...
+    if: >-
+      github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```
DanySK, line 154: why? This should run every time

alchemist-full/PKGBUILD.template
```diff
@@ -0,0 +1,26 @@
+# Maintainer: Marco Sternini <mkarko01@gmail.com>
+# Maintainer: Danilo Pianini <danilo.pianini@unibo.it>
+pkgname=alchemist
+pkgver={%}
+pkgrel=1
+pkgdesc="An extensible simulator for pervasive computing"
+arch=('x86_64')
+url="https://github.com/AlchemistSimulator/Alchemist"
+license=('GPL')
```
DanySK, line 9:
```suggestion
license=('GPL with linking exception')
```

alchemist-full/PKGBUILD.template
```diff
@@ -0,0 +1,26 @@
+# Maintainer: Marco Sternini <mkarko01@gmail.com>
+# Maintainer: Danilo Pianini <danilo.pianini@unibo.it>
+pkgname=alchemist
+pkgver={%}
+pkgrel=1
+pkgdesc="An extensible simulator for pervasive computing"
+arch=('x86_64')
+url="https://github.com/AlchemistSimulator/Alchemist"
```
DanySK, line 8:
```suggestion
url="https://alchemistsimulator.github.io"
```

alchemist-full/build.gradle.kts
```diff
@@ -145,6 +146,42 @@ tasks.register<Exec>("testJpackageOutput") {
...
+    val inputFile = file("${project.projectDir}/PKGBUILD.template")
+    val outputDir = file(rootProject.layout.buildDirectory.map { it.dir("package") }.get().asFile.path)
+    val tokenToReplace = "{%}"
+    var replacementValues: List<String>
+    val testSourceParam = System.getProperty("generatePKGBUILD.testSource", "false")
+    val interceptOutput = ByteArrayOutputStream()
+    standardOutput = interceptOutput
+    doFirst {
+        val version = rootProject.version.toString().substringBefore('-')
+        commandLine("md5sum", "-b", "${rootProject.layout.buildDirectory.get().asFile.resolve("package").path}/${rootProject.name}-$version-1.x86_64.rpm")
```
DanySK, line 161: why do we need `-1` at the end?

DanySK, line 161: we can compute the md5hash in kotlin directly, without going through the command line

release.config.js
```diff
@@ -18,6 +18,8 @@ git -C build/website/ push || exit 3
 RELEASE_ON_CENTRAL="./gradlew uploadKotlinOSSRH uploadKotlinMultiplatform uploadJvm uploadJs release --parallel"
 eval "$RELEASE_ON_CENTRAL" || eval "$RELEASE_ON_CENTRAL" || eval "$RELEASE_ON_CENTRAL" || exit 5
 ./gradlew publishKotlinOSSRHPublicationToGithubRepository --continue || true
+./gradlew generatePKGBUILD
```
DanySK, line 21:
```suggestion
./gradlew generatePKGBUILD || exit <place a number not used elsewhere>
```

DanySK (changes requested):

alchemist-full/build.gradle.kts
Zimbrando: Jpackage outputs packages following that syntax. I suppose it's the release number.

DanySK: There must be a reason, can you investigate? If there is no reason for it to have -1, I'd like to remove it

Zimbrando: It is a naming convention for .rpm packages, .deb packages follow something similar too

> This is tradition and not a restriction, however. Like the version number, the only restriction is that dashes are not allowed. [source](http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html)

I found the jpackage setting, but it seems impossible to disable. 
We can modify it to whatever string we would like

.github/workflows/build-and-deploy.yml
Zimbrando: Ok, I used the configuration of "test-packages". Should I remove the condition in that too?

DanySK: I think so. This check is meant to avoid running the step if there are no secrets. If the step does not requires secrets, it may as well run

DanySK:
The contents of this PR are already in mainline through #3107

Issue #3071, opened 2024-01-22
Unresolved reference: any class inside `graphql` module


AngeloFilaseta:
Importing the `graphql` module in another module by adding it to the `commonMain` as a dependency seems to have no effect.
The problem occurs when importing the module using the statically defined `Project.alchemist(module: String)` function.
The bug won't appear when importing other multiplatform projects (for example, `web-render`).

Only the IntelliJ IDE is able to find the classes and will recognize the import as valid, suggesting a successful import.

Further investigation is required to better understand the problem.

Maybe the `alchemist-graphql` gradle configuration is involved in the issue, since some code generation is involved during the build process.

To reproduce the bug, simply import the `graphql` module in a multiplatform project as a dependency in the `commonMain` block and try to use any class offered. Build will fail upon compiling, here's an example output:

```shell
> Task :alchemist-web-renderer:compileKotlinJvm FAILED
e: file:///home/angelo/AngeloFilaseta/Alchemist/Alchemist/alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/boundary/webui/common/Graph.kt:12:36 Unresolved reference: graphql
e: file:///home/angelo/AngeloFilaseta/Alchemist/Alchemist/alchemist-web-renderer/src/commonMain/kotlin/it/unibo/alchemist/boundary/webui/common/Graph.kt:16:9 Unresolved reference: GraphQLClientFactory
```

DanySK:
This is puzzling. Ca you retry as soon as #3014 gets merged? It's a big refactoring, maybe this one gets accidentally solved.

AngeloFilaseta:
Sure, thanks. If necessary, I can submit a pull request demonstrating the issue, in case this approach doesn't work out.

AngeloFilaseta:
Unfortunately, it didn't solve the problem. It seems to me that the only way to find out what breaks the imports is to remove pieces of configuration until I find out. I'll try to do it next week.

AngeloFilaseta:
This issue will be solved after merging #3253.
The problem was caused by the `application` (or `ktor`) plugin. It seems like applying these plugins have some unknown side effect on the creation of the classpath.

AngeloFilaseta:
I have some additional info:  it seems like merging #3253 did not solve the problem. BTW, the problem is surely related to the `application` plugin.  Adding the `withJava()` directive to all the jvm targets seems to fix the problem. When possible I'll open a PR with the final fix.

[...]

PR #3095, opened 2024-02-01
fix: conflicts between gradle plugins and imports in multiplatform projects

AngeloFilaseta:
This PR should resolve the problem described in issue #3071
It seems like some plugins interfere with the import of classes between multiplatform projects.

Such plugins are:
- `application`
- `ktor` plugin
- `multi-jvm-test`
- `javaQA`

As a solution, I removed the application and ktor plugins from some MP projects. The others are still applied to JVM-only projects. If further investigation is required, I could make an attempt.


AngeloFilaseta:
I know dear robot, you already told me, I will do it, just keep calm

DanySK:
what happened here?

PR #3147, opened 2024-02-21
fix: change molecule extractor to allow the export of short named properties

PaoloPenazzi:


alchemist-loading/src/test/kotlin/it/unibo/alchemist/test/TestExportProperty.kt
```diff
@@ -0,0 +1,36 @@
...
+        Assertions.assertNotNull(loader)
+        val simulation: Simulation<T, P> = loader.getDefault()
+        val exporters = simulation.outputMonitors.filterIsInstance<GlobalExporter<T, P>>().flatMap { it.exporters }
+        exporters.size shouldBe 1
+        val exporter = exporters.first()
+        require(exporter is CSVExporter) {
+            "Invalid exporter of type '${exporter::class.simpleName}'"
+        }
+    }
+})
```
DanySK, line 36: There must be a folder where you put the yaml and it gets tested automatically (maybe `alchemist-loading/src/test/resources/regression/`?)

DanySK (changes requested):

alchemist-loading/src/test/kotlin/it/unibo/alchemist/test/TestExportProperty.kt
PaoloPenazzi: I didn't find anything inside the folder `alchemist-loading`.
Should it look like `./src/test/kotlin/TestWebsiteCodeSnippets.kt`? Which test all the yaml files in a certain folder?

DanySK: I think you can use `TestGuidedTourLoading`, simply add the yaml inside `guidedTour` and see if it works

PaoloPenazzi: Test added

DanySK: And does it work? Namely, if you don't patch, does it fail correctly?

PaoloPenazzi: Yes. 
First, I run the test without changing anything in the code base and it failed (IndexOutOfBound - MoleculeReader.kt - Line 68).
Changing the range in the substring method, as you suggested, made the test pass.

PR #3150, opened 2024-02-22
ci(winget): automatic release to winget packages

Zimbrando:
This PR introduces:

- Automatic release of winget packages through the `wingetcreate `script.
- WiX configuration to automatically set the path when installing a .msi package.
- Fix for the test msi package job to correctly run Alchemist, similar to the other tests.

During testing, I discovered that the current msi packages are not functioning properly, as observed in https://github.com/microsoft/winget-pkgs/pull/138912#issuecomment-1941898707. The windows console flag in jpackage seems to resolve the issue.

Upon PR release, we should update the mentioned winget PR to include the package in the repository. Subsequent runs through winget update will automatically update the version.

The job is triggered when the `successCmd `of the semantic-release plugin is executed. As I did not notice before, the will-release variable should be considered for both the release and the release-winget job.


DanySK:
@Zimbrando can you fix the conflicting files?

Zimbrando:
Now the test is working. The "winget-release" job necessitates an authentication GitHub token for the automatic creation of the pull request for the "winget" package update.

https://github.com/AlchemistSimulator/Alchemist/blob/5b2d775d287558787bb065231a5ad9e9672bfa25/.github/workflows/build-and-deploy.yml#L571-L572

The token needs to be placed in secrets.GIT_AUTH_TOKEN

DanySK:
All releases must be performed by the release job via semantic-release, we cannot have an additional release job.

DanySK:
Ugh, I see, windows needed. Then we should run the actual release step only if release actually released. We need to write an output if so, we cannot reuse the variable computed initially because another commit may have reached the mainline in between and disabled the actual release

.github/workflows/build-and-deploy.yml
```diff
@@ -286,9 +286,13 @@ jobs:
...
           }
-      - name: Print the logs
-        if: always()
-        run: cat install.log
+      - name: Test installation
+        shell: pwsh
+        run: |
+          Import-Module "$env:ChocolateyInstall/helpers/chocolateyInstaller.psm1"
+          refreshenv
+
```
DanySK, line 294:
```suggestion
```

.github/workflows/build-and-deploy.yml
```diff
@@ -521,6 +528,47 @@ jobs:
...
+    runs-on: windows-2022
+    env:
+      packageID: "Unibo.Alchemist"
+    if: >-
+      !github.event.repository.fork
+      && (
+        github.event_name != 'pull_request'
+        || github.event.pull_request.head.repo.full_name == github.repository
+      )
+      && fromJSON(needs.release.outputs.winget-deploy)
```
DanySK, line 549:
```suggestion
    if: fromJSON(needs.release.outputs.winget-deploy)
```

If I understand the workflow correctly, this variable will be true only after a successful release, which in turn means that this condition alone means that we have just released on the main branch.

.github/workflows/build-and-deploy.yml
```diff
@@ -521,6 +528,47 @@ jobs:
...
+    steps:
+      - name: Install winget
+        uses: Cyberboss/install-winget@v1
+      - name: Install wingetcreate
+        shell: pwsh
+        run: Invoke-WebRequest https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
+      - name: Check package availability
+        id: check
+        shell: pwsh
+        continue-on-error: true
```
DanySK, line 559: why continue on error?

DanySK: it seems to be that in case of error the subsequent step would be skipped anyway

DanySK (changes requested):
Almost there!

.github/workflows/build-and-deploy.yml
Zimbrando: I used it as a workaround to avoid the failure of the job the first time the PR gets released (because the winget package is not published yet)

Zimbrando: After thinking about it some more, I believe it can be removed since the release job runs successfully, and even if winget-release fails, the new version is still released, right?

DanySK: The version would be released and then fail, the CI will fail. How to we repair it? Do we have to perform a by-hand release first?

Zimbrando: This PR brings the fix to the MSI package explained before. So in order to release by hand with a correct MSI package, we need first to distribute the package somewhere and put its url in the winget manifest. 

We can split this PR into two separate parts. The first part will address the package fix and the implementation of the test-msi installation. Once this part is completed and released, the next step would be to manually proceed with the winget-release. After that, we can merge this PR correctly.

DanySK: @Zimbrando can you generate the MSI yourself and release it manually, then we merge this?
I don't want to split, everything looks nice

DanySK (approved):

.github/workflows/build-and-deploy.yml
DanySK: let me know when it's done, I'll merge this and then a breaking change PR that I created already. We'll soon know if this works

DanySK: Looks like it's been merged, should we go forward with this one?

Zimbrando: Yes, make sure `secrets.GIT_AUTH_TOKEN` holds a github authentication token, then we can merge this.

DanySK: Can't we simply use `github.token`? Which permissions does it need?

.github/workflows/build-and-deploy.yml
```diff
@@ -521,6 +527,38 @@ jobs:
...
+        shell: pwsh
+        run: |
+          .\wingetcreate.exe show $env:packageID
+      - name: Update package
+        shell: pwsh
+        run: |
+          $version = "${{ needs.ci-preparation.outputs.next-version }}"
+          $packageVersion = if ($version.IndexOf("-") -ne -1) { $version.Substring(0, $version.IndexOf("-")) } else { $version }
+          $installerUrl = "https://github.com/AlchemistSimulator/Alchemist/releases/download/$packageVersion/alchemist-$packageVersion.msi"
+          $gitToken = ${{ secrets.GIT_AUTH_TOKEN }}
```
DanySK, line 559:
```suggestion
          $gitToken = ${{ github.token }}
```

PR #3253, opened 2024-03-25
fix(full): fix import of multiplatform subprojects

AngeloFilaseta:
By merging this PR, `alchemist-full` will become a multiplatform project. This change will allow `alchemist-full` to be able to import dependencies from all the multiplatform subprojects.

In addition, the `run` and `shadowJar` tasks now accurately import the JVM classpath. Previously, there were instances where JVM classes derived from common code were not imported correctly.


alchemist-full/build.gradle.kts
```diff
@@ -33,23 +32,44 @@ plugins {
...
 }
 
 application {
     mainClass.set("it.unibo.alchemist.Alchemist")
 }
 
+/**
+ * Add the runtime classpath of the multiplatform JVM projects to the run task
+ */
+tasks.named("run", JavaExec::class) {
```
DanySK, line 64: Task references by name are very fragile, can't you use something like `tasks.run`?

DanySK (commented):
I'd like to improve the task references, if possible

alchemist-full/build.gradle.kts
AngeloFilaseta: Solved, I also introduced a function to build the `FileCollection`and used it to both configure the `run` and `shadowJar` tasks. Take a look and notify me if something is out of place.

alchemist-full/build.gradle.kts
```diff
@@ -33,23 +32,49 @@ plugins {
...
-    testImplementation(rootProject)
-    testImplementation(alchemist("physics"))
 }
 
+/**
+ * @return a [FileCollection] containing the classpath of the multiplatform JVM projects.
+ * If unspecified, the classpath is not correctly built when including common code from MP dependencies,
+ * resulting in runtime errors.
+ */
+private fun mpClasspath(): FileCollection = listOf(
```
DanySK, line 62: This breaks the configuration avoidance API... I think we need to do a bit of pair programming to see whether this can be solved

AngeloFilaseta: We did it

DanySK (approved):

PR #3386, opened 2024-04-30
feat(graphql)!: add SimulationSurrogate and more support to subscriptions

AngeloFilaseta:
**This is a BREAKING CHANGE since i changed the name of the GraphQLServer (which was an Output Monitor) to GraphQLMonitor**

This PR introduce simpler ways to obtain data using subscriptions in GraphQL.
By creating a `SimulationSurrogate` it is possible to obtain data related to the simulation also using subscriptions and not only using queries.

Please note, this means that it is now possible to do so:

```graphql
subscription Example {
  simulation {
    time
    status
    environment {
      ... other stuff
    }
  }
}
```

however, the following syntax is also valid as a simplification (by omitting the simulation block):

```graphql
subscription Example {
  environment {
    ... other stuff
  }
}
```

If you want the simulation block to always be present I can do that. Personally I like it this way since it allows more customization in shaping the request and the  response.


DanySK:
Please move the development part of the documentation to a separate page, it is not part of the user manual

PR #3683, opened 2024-08-14
feat: add exporter logging nodes positions in the simulation

nicolasfara:


DanySK (commented):
This is good, but if a node is removed and then added, the whole table shifts.

I propose an alternative: we don't specify the expected node count. We read the node count and the highest node id at the first iteration. Then, we check that none of the two changes.

What do you think?

nicolasfara:
From what I understand the `columnNames` property defines the "header" of the exporter, and since it is a `val` how can I read the node count and the highest node ID at the first iteration and then setup the `columnNames`?

In general, I think this exporter is quite "delicate" since if I add a node in the simulation, the exporter will fail badly, right?
Your proposal

Question: if we have 100 nodes, at some point in the simulation we remove the node with ID 99 and after we add a new node, the new one will take the ID 99 or 100? If will take the ID 100 even your strategy will not work since we precompute the `columnNames` with the first-iteration IDs.

DanySK:
My idea is: fail not just if the node count is not the same, but also if the total count is the same, but some node IDs are not identical. New nodes do not acquire the same ID.
Yes it is a bit fragile, and yes I am making it more picky (but failing clearly is good)

nicolasfara:
If I properly understood your idea, I've pushed the updated implementation.
Let me know if it is ok for you.

The CI is failing because of CPD I guess, in particular executing this task `:alchemist-graphql:cpdKotlinCheck`. Is there any mistake from my side?

DanySK:
No, and I could not figure out why the build on master does not fail. It should be fixed now, though.

PR #3844, opened 2024-10-18
fix(graphql): migration to ktor v3

AngeloFilaseta:
`EmbeddedServer` now encapsulates `ApplicationEngine` instead of being a subclass.

Sorry for the other PR, it defaulted to master and I didn't notice.


DanySK:
@AngeloFilaseta, you can target `master` directly. Renovate will auto-close its PR once this one is merged.

AngeloFilaseta:
K, I'll fix it

AngeloFilaseta:
@DanySK I discovered the cause of the issue (the timeout). It is not caused by ktor itself. I already changed the code following the migration guide and everything should work within Alchemist. The problem is with another dependency [graphql-kotlin](https://github.com/ExpediaGroup/graphql-kotlin), which is still not ready to include ktor v3.
For now, the GraphQL server silently fails with `java.lang.NoClassDefFoundError: io/ktor/server/routing/RoutingKt` since it probably expect to find ktor 2 at runtime, but some package structure changed.
Probably, once [this PR](https://github.com/ExpediaGroup/graphql-kotlin/pull/2045) gets included in the new version of the library, everything will work as expected.

So... We wait or what?

DanySK:
Well, I'd say we wait. Set a timer, and let's check back in a month.

DanySK:
you forgot the link :)

DanySK:
This branch has conflicts, can you take a look @AngeloFilaseta

PR #4023, opened 2024-12-18
bug: node of the action and node of the related property do not correspond after spawning

angelacorte:


DanySK:
This "fails almost correctly", in the sense that maybe we can intercept somehow when node properties are being shed around?

PR #4080, opened 2025-01-09
fix(web-renderer): migrate to react  v2025

AngeloFilaseta:
Changing the `Component`  class to `FC` leads to successful  compilation process.

However, __please DON'T force push this to master__, [it could still fail due to the creation of the JS executable](https://youtrack.jetbrains.com/issue/KT-57235).
I don't think react is involved here, the problem is probably related to Kotlin Coroutine in some way, but I need to investigate further. It is possible to forcefully add an additional dependency to solve the problem but this is a sub-optimal solution.

I suggest leaving this PR in auto-[merge|rebase] in case the problem is fixed with other inclusions for now, but I'll keep an eye on it.


DanySK:
enabled auto-rebase + auto-squash-merge.

PR #4096, opened 2025-01-13
feat(loading): create abstract aggregation double exporter

angelacorte:


alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/extractors/AbstractAggregationDoubleExporter.kt
```diff
@@ -0,0 +1,61 @@
...
+
+import it.unibo.alchemist.boundary.ExportFilter
+import it.unibo.alchemist.boundary.exportfilters.CommonFilters
+import it.unibo.alchemist.model.Actionable
+import it.unibo.alchemist.model.Environment
+import it.unibo.alchemist.model.Time
+import it.unibo.alchemist.util.StatUtil
+import org.apache.commons.math3.stat.descriptive.UnivariateStatistic
+
+abstract class AbstractAggregationDoubleExporter(
```
DanySK, line 20:
```suggestion
abstract class AbstractAggregatingDoubleExporter(
```

alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/extractors/AbstractAggregationDoubleExporter.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val filter: String?,
```
DanySK, line 21: pass the ExportFilter with a default, and add a second constructor for the string

DanySK (changes requested):

PR #4099, opened 2025-01-15
feat: add network diameter function based on breadth-first search

angelacorte:


DanySK:
Please write this contribution in Kotlin, thanks

PR #4110, opened 2025-01-17
feat(implementationbase): add utilities to compute shortest paths and network diameters

angelacorte:


alchemist-api/src/main/kotlin/it/unibo/alchemist/model/Environment.kt
```diff
@@ -212,4 +212,37 @@ interface Environment<T, P : Position<out P>> :
      * If node removal is unsupported, it does nothing.
      */
     fun removeNode(node: Node<T>)
+
+    /**
+     * Computes the diameter of the network.
+     * The diameter is the longest shortest path between any two nodes.
```
DanySK, line 218: It does not work in the case of segmented networks. We have two possibilities:
1. return `NaN` if the network is segmented
2. return a `List<Int>` with one diameter per network segment (better, return a List<Set<Node<T>, Int>, mapping each subset of nodes to its diameter)
3. require a `Node` as input and compute the diameter of that segment

DanySK (approved):

PR #4115, opened 2025-01-19
feat: introducing multi-platform UI with Compose

Tbaile:
Introducing new project that tries to give Alchemist a UI that is completely multi-platform across desktop apps and web.
This is an introductory pull requests that adds the project and the initial dependencies.


alchemist-composeui/build.gradle.kts
```diff
@@ -0,0 +1,92 @@
...
+import Libs.alchemist
+import de.aaschmid.gradle.plugins.cpd.Cpd
+import io.gitlab.arturbosch.detekt.Detekt
+import org.jetbrains.kotlin.gradle.ExperimentalWasmDsl
+import org.jetbrains.kotlin.gradle.targets.js.webpack.KotlinWebpackConfig
+import java.io.File.separator
+
+plugins {
+    id("kotlin-multiplatform-convention") apply false
+    kotlin("multiplatform")
```
DanySK, line 19:
```suggestion
    id("kotlin-multiplatform-convention")
```

If there are changes required to the convention, then we apply them to the convention

alchemist-composeui/build.gradle.kts
```diff
@@ -0,0 +1,92 @@
...
+                    }
+            }
+        }
+        binaries.executable()
+    }
+
+    jvm {
+        withJava()
+    }
+
```
DanySK, line 48:
```suggestion
```

alchemist-composeui/build.gradle.kts
```diff
@@ -0,0 +1,92 @@
...
+                implementation(libs.androidx.lifecycle.runtime.compose)
+            }
+        }
+    }
+}
+
+publishing.publications.withType<MavenPublication>().configureEach {
+    pom {
+        contributors {
+            contributor {
```
DanySK, line 74: You can uplift yourself as developer

DanySK (changes requested):
let's make sure the build is clean

alchemist-composeui/build.gradle.kts
Tbaile: While the convention is fine as is, it requires the `js` target which is not available in this package. A possible solution is to being able to customize the convention, this needs then enough defaults so that the convention does not break everything and can be customized when needed. If that works for you I'll look into that.

Tbaile: By changing the section from the `multiplatform` from this:
```kotlin
js {
    browser()
    nodejs()
}
```

to this:
```kotlin
// adding js only when directory `jsMain` exists
if (projectDir.resolve("src/jsMain").exists()) {
    js {
        browser()
        nodejs()
    }
}
```
It's possible to achieve this behavior, not sure it's most optimal one (but might be effective in the long run, just checking for the target to build before importing the target)

DanySK: Wait, why don't we have js? Isn't Compose supposed to run also in browser? This is also our main goal: an in-browser interface

Tbaile: It runs on web, but not using the `js` target, instead using the `wasmJs` one. This is the only way to use the cross-platform interface: https://kotlinlang.org/docs/wasm-overview.html

DanySK: Oh, I see. I'd proceed as follows:

1. create `withJS()` extension method in the convention, running `js { browser() nodejs() }`
2. create `withWasmJS()`  extension method in the convention, running `js { browser() nodejs() }`
3. enable only JVM inside the convention
4. update all other modules using the convention to call `withJS()`

I would avoid fancy autodetection of folders. It can get very surprising, and the logics is reversed (if I have JS, then I may have the folder, not vice versa). I could (and I do) generate js for purely kotlin subprojects

alchemist-composeui/build.gradle.kts
```diff
@@ -0,0 +1,92 @@
...
+                email.set("tommaso.bailetti@studio.unibo.it")
+            }
+        }
+    }
+}
+
+tasks {
+    fun PatternFilterable.excludeGenerated() = exclude { "build${separator}generated" in it.file.absolutePath }
+    withType<Detekt>().configureEach { excludeGenerated() }
+    ktlint { filter { excludeGenerated() } }
```
DanySK, line 85: I think this can be moved in the convention

alchemist-composeui/src/wasmJsMain/kotlin/it/unibo/alchemist/boundary/composeui/Main.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.compose.ui.window.ComposeViewport
+import kotlinx.browser.document
+
+/**
+ * Main entry point that loads the compose generic application in the browser.
+ * Even if not documented, this loading process needs an index.html file to load up the produced JavaScript.
+ */
+@OptIn(ExperimentalComposeUiApi::class)
+fun main() {
+    ComposeViewport(document.body!!) {
```
DanySK, line 22:
```suggestion
    ComposeViewport(checkNotNull(document.body) { "Reasonable error message" }) {
```

gradle/libs.versions.toml
```diff
@@ -122,6 +124,7 @@ spotbugs-annotations = "com.github.spotbugs:spotbugs-annotations:4.9.0"
 redux-kotlin-threadsafe = "org.reduxkotlin:redux-kotlin-threadsafe:0.6.1"
 svgsalamander = "guru.nidi.com.kitfox:svgSalamander:1.1.3"
 trove4j = "net.sf.trove4j:trove4j:3.0.3"
+androidx-lifecycle-runtime-compose = { group = "org.jetbrains.androidx.lifecycle", name = "lifecycle-runtime-compose", version.ref = "androidx-lifecycle" }
```
DanySK, line 127: please use lexicographic ordering

gradle/libs.versions.toml
```diff
@@ -167,3 +170,5 @@ scalafmt = "cz.alenkacz.gradle.scalafmt:1.16.2"
 shadowJar = "com.github.johnrengelman.shadow:8.1.1"
 taskTree = "com.dorongold.task-tree:4.0.0"
 jpackage = "org.panteleyev.jpackageplugin:1.6.0"
+compose-multiplatform = { id = "org.jetbrains.compose", version.ref = "compose-multiplatform" }
+compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
```
DanySK, line 174: lexicographic ordering

settings.gradle.kts
```diff
@@ -40,6 +40,7 @@ include(
     "alchemist-ui-tooling",
     "alchemist-swingui",
     "alchemist-web-renderer",
+    "alchemist-composeui",
```
DanySK, line 43: lexicographic order

alchemist-composeui/build.gradle.kts
DanySK: Actually, @nicolasfara run a check, and we can generate both js and wasmJs: https://github.com/nicolasfara/rstcovers-task-manager/blob/master/frontend/src/jsMain/kotlin/it/nicolasfarabegoli/rstcovers/Main.kt

Tbaile: Nice to know new undocumented things of compose :sweat_smile: 
Will work on this on the weekend.

Tbaile: Me and @AngeloFilaseta delved in the why @nicolasfara's code worked despite ComposeUI missing completely in the JS multiplatform documentation. It appears that putting the Compose UI implementation inside the JS is not supported and breaks the web application entirely.

Only UI coming from `org.jetbrains.compose.web.` appears to be rendered successfully and without issues. This is possibly due to Jetbrains bad wording in choosing the different product names.

I've still pushed a POC if you're willing to try it. (but will 100% break when run with `./gradlew :alchemist-composeui:jsBrowserDevelopmentRun`).

alchemist-composeui/build.gradle.kts
Tbaile: Couldn't manage to do it for now. @AngeloFilaseta it's going to take a closer look at it in the following days.

alchemist-composeui/build.gradle.kts
DanySK: If I understand correctly -- js compiles successfully but doesn't render correctly, is this correct?

Tbaile: Yes, it compiles the JS and serves it correctly, however when ran by browser an error is thrown and it appears irrecoverable

DanySK: Then keep the js target enabled. We'll default on WASM, but there's no reason no to generate js, too

DanySK (approved):

PR #4401, opened 2025-04-19
feat: add SystemEnvVariable to load environment variable

cric96:


alchemist-loading/src/main/kotlin/it/unibo/alchemist/boundary/variables/SystemEnvVariable.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.util.stream.Stream
+
+/**
+ * A variable that retrieves its value from a system environment variable.
+ *
+ * @param name the name of the environment variable
+ * @param defaultValue the default value to return if the environment variable is not set
+ */
+class SystemEnvVariable(private val name: String, private val defaultValue: Serializable?) : Variable<Serializable> {
+    constructor(name: String) : this(name, null)
```
DanySK, line 23:
```suggestion
/**
 * A variable that retrieves its value from a system environment variable.
 *
 * @param name the name of the environment variable
 * @param defaultValue the default value to return if the environment variable is not set.
 * If no default is provided and the environment variable is unset, this variable will throw an `IllegalStateException`
 */
@JvmOverloads
class SystemEnvVariable(private val name: String, private val defaultValue: Serializable?) : Variable<Serializable> {
```

Prefer default parameters + `@JvmOverloads` over constructor overloading

DanySK (changes requested):
I'll make a couple of changes myself, it's faster :)

DanySK (approved):

PR #5556, opened 2026-08-17
feat(geospatial): add alchemist-geospatial module for real-world climate data

Wanes01:


alchemist-geospatial/src/main/kotlin/it/unibo/alchemist/boundary/acquisition/utility/Integrity.kt
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Computes the MD5 digest of [file] as a lowercase hex string,
+ * streaming the file through the digest so arbitrarily large files
+ * are never fully held in memory.
+ *
+ * @param file the file to digest.
+ * @return the MD5 digest, as a lowercase hex string.
+ */
+internal fun md5Hex(file: Path): String {
+    val digest = MessageDigest.getInstance("MD5")
```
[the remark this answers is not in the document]
Wanes01: Ciò che segnala SonarCloud ("weak hash algorithm" su `md5Hex`) è un falso positivo in questo contesto: l'MD5 viene usato esclusivamente per verificare l'integrità dei file scaricati confrontandolo con il checksum fornito dall'API Copernicus (CDS/ADS/EWDS), che riporta il digest MD5 prima del download e impone questo specifico algoritmo.

Ho aggiunto una direttiva "NOSONAR" nel codice, ma il check continua a fallire.

alchemist-full/build.gradle.kts
```diff
@@ -323,3 +323,13 @@ private val Provider<RegularFile>.fileProperty get(): RegularFileProperty = obje
...
+
+configurations.all {
+    resolutionStrategy {
+        eachDependency {
+            if (requested.group == "com.google.protobuf" && requested.name == "protobuf-java") {
+                useVersion("3.12.2")
+            }
+        }
+    }
+}
```
DanySK: Why is this there? This induces a big problem with the entire dependency management of the simulator.

alchemist-geospatial/src/main/kotlin/it/unibo/alchemist/boundary/acquisition/utility/Archives.kt
```diff
@@ -0,0 +1,73 @@
...
+ * @return `true` if [archive] was a valid ZIP (its entries were extracted), `false` if it is not a
+ * ZIP and was left untouched.
+ * @throws IllegalStateException on a basename collision between two entries.
+ */
+private fun extractArchive(archive: Path, dir: Path): Boolean = try {
+    ZipFile(archive.toFile()).use { zip ->
+        for (entry in zip.entries()) {
+            // nothing to do if the entry is directory.
+            if (entry.isDirectory) continue
+
```
DanySK: avoid inconsistent vertical spacing. No empty lines within methods. This applies throughout the codebase

alchemist-geospatial/src/main/kotlin/it/unibo/alchemist/boundary/acquisition/utility/Archives.kt
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (C) 2010-2026, Danilo Pianini and contributors
+ * listed, for each module, in the respective subproject's build.gradle.kts file.
+ *
+ * This file is part of Alchemist, and is distributed under the terms of the
+ * GNU General Public License, with a linking exception,
+ * as described in the file LICENSE in the Alchemist distribution's top directory.
+ */
+
+package it.unibo.alchemist.boundary.acquisition.utility
```
DanySK: Why "acquisition"?
Also, in the simulator, we have a `utils` package. This file may be relocated in `it.unibo.alchemist.boundary.utils`

Also, this looks like a very general zip utility, why do we need to unpack zips? Can't we read them directly with the appropriate stream?

alchemist-maps/build.gradle.kts
```diff
@@ -42,6 +42,16 @@ dependencies {
     testRuntimeOnly(incarnation("protelis"))
 }
 
+configurations.all {
+    resolutionStrategy {
+        eachDependency {
+            if (requested.group == "com.google.protobuf" && requested.name == "protobuf-java") {
+                useVersion("3.12.2")
```
DanySK: This is also a big issue

build.gradle.kts
```diff
@@ -56,6 +56,7 @@ allprojects {
     repositories {
         google()
         mavenCentral()
+        maven("https://artifacts.unidata.ucar.edu/repository/unidata-all/")
```
DanySK: Nope, we can't. Why do we need this? I'd rather republish this on Central. We cannot depend on artifacts that are not available on Maven Central.

DanySK (commented):
There are several critical issues that must be solved before I go ahead with the review. See comments.

alchemist-full/build.gradle.kts
Wanes01: There is a conflict between `osmosis-osm-binary`, which requires protobuf <= 3.25.5, and `cdm-core` (used to read formats such as NetCDF), which brings protobuf version 4.31.1.
I wasn't sure how to resolve the issue, so I resolved the conflict this way in both alchemist-maps and alchemist-full.

alchemist-geospatial/src/main/kotlin/it/unibo/alchemist/boundary/acquisition/utility/Archives.kt
Wanes01: I'm on it.

build.gradle.kts
Wanes01: The `cdm-core` used in the project requires version >= 5.x, which is not published on Central. There is an old artifact on Central with version 4.3.22; I tried using that one, but it had problems reading the latest version of NetCDF (NetCDF4).

alchemist-geospatial/src/main/kotlin/it/unibo/alchemist/boundary/acquisition/utility/Archives.kt
Wanes01: I didn't know there was a subpackage for utilities; I'll move all the utilities there.
Regarding the stream: there isn't a generic stream that cdm-core can use directly. cdm-core does not accept a raw InputStream when opening a NetCDF or GRIB file.

Wanes01: PS: “acquisition” because the purpose of the components within this subpackage is to acquire resources, whether they reside on the filesystem or on the internet.

build.gradle.kts
DanySK: I fixed this. Took a while

alchemist-full/build.gradle.kts
DanySK: Ho do I trigger the failure from `osmosis-osm-binary`? I forced protobuf 4.35.1 in alchemist-maps and all tests pass.

Wanes01: The problem occurred when trying to use a `CopernicusLayer` while simultaneously passing a `.pbf` extract to `OSMEnvironment`. Now the problem no longer occurs, but I don't exactly understand why. I looked into it a bit but couldn’t find the reason for the fix. In any case, I added a `StringSpec` that should fail if the problem reoccurs. For now, it builds and runs without issues, so I’ve removed this snippet from both here and from the `build.gradle.kts` file in `alchemist-maps`.

alchemist-geospatial/src/main/kotlin/it/unibo/alchemist/boundary/acquisition/utility/Archives.kt
Wanes01: I moved the utility files to the `utils` subpackage. Is this organization okay?

alchemist-maps/build.gradle.kts
Wanes01: See the comment under `alchemist-full/build.gradle.kts`

